PR #1, opened 2024-01-02
Meson updates for monorepo

HeliumAnt:


external/sources/luabind-0.7.1/meson.build
```diff
@@ -20,17 +20,12 @@ elif cxx.get_argument_syntax() == 'msvc'
...
-  luabind = static_library('luabind071', luabind_src, dependencies: luabind_dependencies, include_directories:luabind_include, cpp_args:cpp_args+ luabind_args, override_options: ['warning_level=0']+ cpp_options)
-else
-  luabind = []
-  if get_option('debug')
-    deps = cxx.find_library('luabind-debug' , dirs: meson.current_source_dir()/'_Bin')
-  else
-    deps = cxx.find_library('luabind-release', dirs: meson.current_source_dir()/'_Bin')
-  endif
+if (not get_option('debug')) or get_option('debug_type') == 'release'
+  luabind_args += ['-DLUABIND_NO_ERROR_CHECKING=1']
```
Causeless, line 24: Worth noting that we might go back on this. It gives a significant performance gain, but it also means lua errors can just force-crash the game. Need to consider the best way of handling this...

Causeless (approved):

Issue #5, opened 2024-01-05
The Browncoat Update: Some way to disable HUD like control mode display, player gold amount, etc.


pawnishoovy:
Some FrameMan function or other is needed to disable HUD.

Whether this is piecemeal or just a full on/off toggle, and whether things like the buy menu would fall into such a toggle, I don't know. But we at least need the basic HUD togglable for the cinematic pieces in the upcoming mission.

PR #7, opened 2024-01-05
Disable HUD via Script

Causeless:
Fixes https://github.com/cortex-command-community/Cortex-Command-Community-Project/issues/5


Activities/GameActivity.cpp
```diff
@@ -1430,55 +1430,50 @@ void GameActivity::Update()
                 m_ControlledActor[player] = 0;
         }
 
-        ///////////////////////////////////////////
         // Player-commanded actor switching
+        if (m_ViewState[player] != ViewState::Observe) {
```
Causeless, line 1434: This looks like a bit change but it's just adding this if against observe here - this is to stop player controlling actors when forced into observe state

PR #13, opened 2024-01-10
Linux arm64 compilation support

janisozaur:
This PR adds **compilation** support for ARM64-based Linux.

There were two issues:
1. fmod was only added as x86_64 binary
2. RakNet required minor adjustment for architecture-specific include.

I took the liberty to update fmod from 2.2.13 to latest version 2.2.20, see https://www.fmod.com/docs/2.02/api/welcome-revision-history.html

I was not able to locate RakNet's origin to validate if the fix applied is available upstream.

Do note this is my first PR, please instruct me if there are any steps I need to take before this can get merged.


Causeless (approved):

Causeless:
Nice work!

PR #15, opened 2024-01-14
Add filesystem functions

MyNameIsTrez:
New `LuaMan` Lua I/O functions `DirectoryCreate`, `FileExists`, `DirectoryExists`, `FileRemove`, `DirectoryRemove`, `FileRename`, `DirectoryRename` and `IsValidModulePath`.

I tried making these case-sensitive for ages, but eventually realized that while possible for some like `FileExists`, it is impossible for `CreateDirectory()` and others. So I decided to let Linux emulate Windows its case-insensitivity. This still achieves my primary goal of having the strong guarantee that any mod that works on Windows will always work on Linux, and vice versa.

I wrote a 238 line long file of Lua tests, which are absolutely necessary in order to make sure stuff isn't broken with future changes. I have made sure that both Windows and Ubuntu pass all of these tests.
So if someone wants to touch these functions in the future, please use [the mod](https://github.com/cortex-command-community/Cortex-Command-Community-Project/files/13933291/filesystem.rte.zip), which runs the test as a global script. When `End of Filesystem:StartScript()` is printed, you know all the test assertions passed.


Source/Managers/LuaMan.cpp
```diff
@@ -923,40 +941,116 @@ namespace RTE {
 
 /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
-	const std::vector<std::string> * LuaMan::DirectoryList(const std::string &filePath) {
+	// TODO: Ask Causeless whether this is an appropriate spot for this method
```
MyNameIsTrez, line 944: @Causeless

Source/Managers/LuaMan.cpp
```diff
@@ -923,40 +941,116 @@ namespace RTE {
...
+	std::string LuaMan::GetCaseInsensitiveFullPath(const std::string &fullPath) {
```
MyNameIsTrez, line 945: @Causeless Thoughts on this TODO?

Causeless: Move it to RTETools - it can go alongside StringsEqualCaseInsensitive()

Source/Managers/LuaMan.h
```diff
@@ -8,6 +8,8 @@
 
 #include "BS_thread_pool.hpp"
 
+#include <optional>
```
Causeless, line 11: Is this necessary? Where are we using this?

MyNameIsTrez: Good catch! Removed that vistigial #include

Source/Managers/LuaMan.cpp
MyNameIsTrez: Done

Causeless (dismissed):
Perf concerns about how many io operations are required for getting the proper path, but ah well, not much choice I guess

PR #22, opened 2024-01-19
clang-format in CI

janisozaur:


janisozaur:
Once this is merged, I will send a PR that adds formatting commit to `.git-blame-ignore-revs` file. See https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/

Causeless (approved):

PR #27, opened 2024-01-21
Fix some GCC warnings

janisozaur:


janisozaur:
I plan to continue fixing those warnings, but would like to have the first batch reviewed/merged.

This reduces `-Wsign-compare` from 1014 to 179, eliminates `-Wreorder` and `-Wdangling-else` with GCC 11.

Causeless (approved):

PR #34, opened 2024-01-23
un-PCH and use ccache in CI

janisozaur:
ccache greatly improves recompilation turnaround by skipping the compilation and reusing already compiled artifacts.

Some compiler options, like PCH, affect cacheability. Ccache has some support for using PCH, called `sloppiness`, but I prefer not to use it, as I fear it could lead to non-trivial issues when the PCH changes, so I implemented un-PCH mode.

msbuild was refactored to have some common properties extracted to a common file.

There's also the issue of old macos runner being used. It's too old to have ccache precompiled binaries available, I may try updating the runner in the future.

You can control use of PCH with `-Duse_pch=` and set it to
* `auto` (default), will search for `ccache` executable and if found, will disable PCH. When no `ccache` is found, PCHs will be used
* `enabled` to force use of PCH
* `disabled` to force disable of PCH


RTEA.vcxproj
```diff
@@ -155,13 +155,14 @@
...
-      <PreprocessorDefinitions>_ITERATOR_DEBUG_LEVEL=0;WIN32_LEAN_AND_MEAN;ALLEGRO_STATICLINK;ALLEGRO_NO_STD_HEADERS;ALLEGRO_NO_MAGIC_MAIN;_CRT_SECURE_NO_DEPRECATE;ZLIB_WINAPI;ZLIB_COMPAT;_WINDOWS;WIN32;DEBUG_BUILD;DEBUGMODE;TARGET_MACHINE_X86;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
Causeless, line 164: I think removing all of these doesn't work with normal non-CI builds, surely? We don't typically use meson for normal non-CI Visual Studio builds, so I don't think the pre processor flags are added.

However maybe I am misunderstanding and the defines are set elsewhere?

Causeless (changes requested):

RTEA.vcxproj
Causeless: Nevermind, duh, common props. All good!

Causeless (approved):

PR #37, opened 2024-01-24
GL updates

janisozaur:


Data/Base.rte/Shaders/ScreenBlit.frag
```diff
@@ -20,5 +20,5 @@ void main() {
 	vec4 guiColor = texture2DAA(rteGUITexture, textureUV);
 	float guiSolid = step(0.00000001, guiColor.r + guiColor.g + guiColor.b);
 	float blendRatio = max(guiColor.a, guiSolid);
-	FragColor = (texture2DAA(rteTexture, textureUV) * (1 - blendRatio)) + guiColor * blendRatio;
+	FragColor = (texture2DAA(rteTexture, textureUV) * (1.0 - blendRatio)) + guiColor * blendRatio;
```
Causeless, line 23: Why not 1.0F?

Data/Base.rte/Shaders/Blit8.frag
```diff
@@ -6,9 +6,9 @@ in vec2 textureUV;
 out vec4 FragColor;
 
 uniform sampler2D rteTexture;
-uniform sampler1D rtePalette;
+uniform sampler2D rtePalette;
 
 void main() {
 	float colorIndex = texture(rteTexture, vec2(textureUV.x, -textureUV.y)).r;
-	FragColor = texture(rtePalette, colorIndex);
+	FragColor = texture(rtePalette, vec2(colorIndex, 0.f));
```
Causeless, line 13: We typically format floats always including the 0 and capitalised F, so 0.0F

Data/Base.rte/Shaders/Blit8.frag
```diff
@@ -6,9 +6,9 @@ in vec2 textureUV;
 out vec4 FragColor;
 
 uniform sampler2D rteTexture;
-uniform sampler1D rtePalette;
```
Causeless, line 9: Why? Are 1d textures unsupported on some platforms or something?

Causeless (changes requested):
Couple comments, but I don't know much about shader programming so feel free to swat em away 🫡

Data/Base.rte/Shaders/Blit8.frag
HeliumAnt: I'm guessing for gles support? Anything else should have support for tex1d, though I'm guessing driver implementation is probably mostly same as this here.

Data/Base.rte/Shaders/ScreenBlit.frag
HeliumAnt: glsl implicitly uses float for everything, specfying f is redundant

Data/Base.rte/Shaders/Blit8.frag
janisozaur: Yes

Causeless:
Does implementing GLES at runtime give us any real benefit? Does that just mean we can query for the best API at startup instead of  (necessarily) providing different binaries, or is there some other reason?

Causeless (approved):

PR #40, opened 2024-01-24
OpenGL ES

janisozaur:
This is #37 with actual gles commit on top for discussion.

The code was tested on several machines and in some cases there were issues with rendering text. The platforms affected were laptop nvidia and rpi5/vc7 and the issue observed was exactly the same. While nvidia was easily addressed by adding `highp`, the same issue on rpi5 persists. Interestingly, on another ARM-based device with RK3588 and Mali GPU this was not observed. I suspect this might be rpi quirk and will see if I manage to debug this in any way.

Any help appreciated. I attach screenshots of the issue.
cc_gles_4x
cc_gles_1x


Causeless:
I'm not sure this is specifically a text rendering issue. As far as I know, text in CC is just using a spritesheet and treated no differently to any other graphics in the game. It's relatively complete, with kerning implemented, but the actual font rendering is nothing beyond a blit- there's no vector representation of the text characters.

I suspect that if there's an issue here, it's actually some sort of filtering or interpolation applied over the entire screen, and it's just most obvious with text.

Causeless:
If you look at the left side of the picture, there looks to be a lot of stray white pixels that I don't recall seeing before. Almost as if the text pixels were placed in the wrong place.

HeliumAnt:
Looks like the palette is definitely getting corrupted but probably everything else that's uploaded as well, considering how that text looks

janisozaur:
If it was the palette, wouldn't just the colors be shifted? The way it works now is there are vertical stripes (not really, but it looks like this when the text expands) on the screen that are somehow filtered from getting proper rendering.

HeliumAnt:
ok not the palette, that would have to create waaay more grey spots, since the greyed out colors on the background occur in a lot more places. Also this has to happen somewhere in software drawing, since otherwise I'd expect more issues on non text elements and more general artifacts, especially *any* artifacts on the scrolling background. Looks like it may just be things drawn by the putpixel methods (scene color)?
(or the assets are already corrupted, which seems unlikely since the letters all fail differently).

Causeless:
Hey, I can't really merge this up-to-date as I don't know what I'm doing, but there was some fixes in dev where some pallete indices were messed up.

@janisozaur can you try merging uip to dev and seeing if this is still an issue?

PR #41, opened 2024-01-25
Mark GUIControl destructor as virtual

janisozaur:
Previously GUIControl destructor was not virtual and did not run any of derived classes destructors. Now that it is and derived destructors are ran, they actually double free due to `Destroy` has already been called on the objects by `GUIControlManager::Clear`. They were also inconsistent with some classes having them and some not, so I unified the approach and removed the destructors in `GUIControl`-derived classes that only called `Destroy`.


Source/GUI/GUISkin.h
```diff
@@ -11,10 +11,6 @@ namespace RTE {
 		/// @param Screen GUIScreen Interface.
 		explicit GUISkin(GUIScreen* Screen);
 
-		/// Destructor method used to free a GUISkin object in system
-		/// memory.
-		~GUISkin();
```
Causeless, line 16: Isn't this going to cause leaked memory? Or is destroy called somewhere else externally?

I think it'd be more consistent in that case if we make Destroy private instead and just let RAII clean it up in the destructor, instead of requiring the caller to manually Destroy() the skin.

Source/GUI/GUISound.h
```diff
@@ -22,9 +22,6 @@ namespace RTE {
 #pragma region Destruction
 		/// Destroys and resets (through Clear()) the GUISound object.
 		void Destroy() { Clear(); }
-
-		/// Destructor method used to clean up a GUISound object before deletion from system memory.
-		~GUISound();
```
Causeless, line 27: Same thoughts as above

Source/GUI/GUICollectionBox.h
```diff
@@ -24,9 +24,6 @@ namespace RTE {
 		/// @param Manager GUIManager, GUIControlManager.
 		GUICollectionBox(GUIManager* Manager, GUIControlManager* ControlManager);
 
-		/// Destructor method used to clean up this before deletion from memory.
-		~GUICollectionBox();
```
Causeless, line 28: As with other comments

Causeless (changes requested):

Source/GUI/GUISound.h
Causeless: Oh - I guess this one isn't relevant as Destroy() just clears

Causeless:
Resolved all comments after reading code more closely. This looks good given the circumstances.

Existing code is trash but I guess it's what we've gotta deal with :/

PR #42, opened 2024-01-25
Only perform activity actions when activity is running

janisozaur:
When game starts, there is no activity running yet, but when loading
data through `g_PresetMan.LoadAllDataModules()` call in `main()`, this
will get executed and call a member function on a nullptr.


Causeless (changes requested):
Hmm, I'm not sure if this is the best way of doing this. Generally speaking we shouldn't be doing anything in MovableMan if no activity is running: it'll be in an invalid state. This change looks like it might makes things even more of an invalid state.

We should probably assert if any operations are performed in MovableMan while an activity is not running. 

Anyways, from what what I can see, DataModule shouldn't interact with MovableMan at all. May I ask how you are reproducing this? I'm assuming through some mod?

I am guessing it's from a ScriptPath, but I have some thoughts on removing ScriptPath from DataModule entirely. It's not really useful in any way anymore now that Lua scripts are properly sandboxed.

janisozaur:
So `RTEAssert` then?

Causeless:
Yes, that would make more sense. May I ask how you are reproducing this crash though? Is it some mod that causes this?

janisozaur:
I changed it to `RTEAssert` then

Causeless (approved):

PR #44, opened 2024-01-29
Let filesystem functions be slightly less strict

MyNameIsTrez:
So the timeline of events is that:
1. My original implementation of `DirectoryList()` forgot the logic that disallows escaping the game's directory.
2. I noticed this, and panicked so hard that I let my recent filesystem functions PR disallow mods from doing anything outside of their directory, just to be safe.
3. Today I tried to actually use my new filesystem functions in ModMod, and quickly noticed that my mod entirely broke, since it wasn't able to iterate over the entries in `Data/` or `Mods/` anymore. I also realized that loads of functionality I am still planning to add, like being able to create or rename mod directories in-game, was now entirely impossible. But there isn't anything fundamentally unsafe with these things!

So, this PR finds a good middle ground, where the game's directory still can't be escaped, and shenanigans like editing the game executable's bytes still isn't allowed, but safe operations like iterating over/creating/moving directories around is now allowed.

I also got rid of the `IsValidModulePath()` binding, since I realized that I don't think I'd ever even find a use for it myself, as you'd always want to either specifically use either `FileExists()` or `DirectoryExists()` instead in practice. A binding for it could always be added back later, if someone finds a use for it.

### DIFF
--- a/Source/Managers/LuaMan.cpp
+++ b/Source/Managers/LuaMan.cpp
@@ -78,7 +78,6 @@ void LuaStateWrapper::Initialize() {
 	                             .def("GetFileList", &LuaStateWrapper::FileList, luabind::adopt(luabind::return_value) + luabind::return_stl_iterator)
 	                             .def("FileExists", &LuaStateWrapper::FileExists)
 	                             .def("DirectoryExists", &LuaStateWrapper::DirectoryExists)
-	                             .def("IsValidModulePath", &LuaStateWrapper::IsValidModulePath)
 	                             .def("FileOpen", &LuaStateWrapper::FileOpen)
 	                             .def("FileClose", &LuaStateWrapper::FileClose)
 	                             .def("FileRemove", &LuaStateWrapper::FileRemove)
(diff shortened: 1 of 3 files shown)


Source/Managers/LuaMan.cpp
```diff
@@ -885,17 +883,17 @@ LuaMan::~LuaMan() {
 }
 
 const std::vector<std::string>* LuaMan::DirectoryList(const std::string& path) {
-	std::string fullPath = System::GetWorkingDirectory() + g_PresetMan.GetFullModulePath(path);
+	std::string fullPath = System::GetWorkingDirectory() + path;
 	auto* directoryPaths = new std::vector<std::string>();
 
-	if (IsValidModulePath(fullPath)) {
+	if (fullPath.find("..") == std::string::npos) {
```
Causeless, line 889: Doesn't this still allow absolute directory access? I.e C:/system32

Causeless (approved):

Source/Managers/LuaMan.cpp
MyNameIsTrez: Nah, cause `fullPath` is `std::string fullPath = System::GetWorkingDirectory() + path;`, so it is forced to start with the game's directory.

PR #45, opened 2024-01-30
Fix signedness of some members 

HeliumAnt:
Fixes some bugs regarding signedness (namely limbpath), and makes places i could find use size_t for indexing instead of unsigned, and tries to remove unsigned from some other contexts


Causeless (approved):
Thank you for addressing this!

PR #47, opened 2024-02-01
Added line, circle and box variants of DislodgePixel to SceneMan

comradeshook:
As per title, added the following functions to SceneMan:
``DislodgePixelLine(Vector startPos, Vector ray, int skip, bool deletePixels = false)`` - Calls DislodgePixel() on all pixels in a line (density depending on the skip value).
``DislodgePixelCircle(Vector centre, float radius, bool deletePixels = false)`` - Calls DislodgePixel() on all pixels in a circle.
``DislodgePixelBox(Vector upperLeftCorner, Vector lowerRightCorner, bool deletePixels = false)`` - Calls DislodgePixel() on all pixels in a box.

All of them share that they return a list of the dislodged pixels, and that the pixels will be set to delete if the last bool is set to true. If omitted, it defaults to false. The reasoning behind this is that mass-calling DislodgePixel() in C++ is much faster than doing the same in Lua, and may have use for modders.

Additionally, added a bool variant of DislodgePixel, ``DislodgePixel(float posX, float posY, bool deletePixel)``, which deletes the dislodged pixel immediately. The original function is unchanged in order to avoid breaking existing code.


Causeless:
Please update CHANGELOG.md too if you can!

comradeshook:
> Please update CHANGELOG.md too if you can!

Done! Also fixed scene wrapping and the for-loops using floats. :v

Causeless (approved):

Causeless (dismissed):

Causeless (approved):

PR #52, opened 2024-02-04
Pre6 more tweaks and changelog

pawnishoovy:
data-related changelog, fix dropship AI quirk, clean up a bunch of code so there's no leftover prints and whatnot

i hope this is the last PR of this kind we'll need


Causeless (approved):
Looks good. As aforementioned I'll be picking out some of the more data oriented changelog stuff for release notes instead of the changelog, but that can be done post-merge

PR #53, opened 2024-02-04
Exposed MOPixel Color to Lua, made DislodgePixel return MOPixels

comradeshook:
Enabled Color changes of `MOPixels` on the fly with Lua, using two new methods to specifically get and set the index of the MOPixel's color directly. Additionally, changed `DislodgePixel` and all of its new derivatives to return `MOPixel` instead of `MovableObject`, because what else would it possibly return?

### DIFF
--- a/Source/Lua/LuaBindingsEntities.cpp
+++ b/Source/Lua/LuaBindingsEntities.cpp
@@ -763,7 +763,8 @@ LuaBindingRegisterFunctionDefinitionForType(EntityLuaBindings, MOPixel) {
 	return ConcreteTypeLuaClassDefinition(MOPixel, MovableObject)
 
 	    .property("TrailLength", &MOPixel::GetTrailLength, &MOPixel::SetTrailLength)
-	    .property("Staininess", &MOPixel::GetStaininess, &MOPixel::SetStaininess);
+	    .property("Staininess", &MOPixel::GetStaininess, &MOPixel::SetStaininess)
+	    .property("Color", &MOPixel::GetColorIndex, &MOPixel::SetColorIndex);
 }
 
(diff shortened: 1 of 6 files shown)


Source/Lua/LuaBindingsEntities.cpp
```diff
@@ -763,7 +763,8 @@ LuaBindingRegisterFunctionDefinitionForType(EntityLuaBindings, MOPixel) {
 	return ConcreteTypeLuaClassDefinition(MOPixel, MovableObject)
 
 	    .property("TrailLength", &MOPixel::GetTrailLength, &MOPixel::SetTrailLength)
-	    .property("Staininess", &MOPixel::GetStaininess, &MOPixel::SetStaininess);
+	    .property("Staininess", &MOPixel::GetStaininess, &MOPixel::SetStaininess)
+	    .property("Color", &MOPixel::GetColorIndex, &MOPixel::SetColorIndex);
```
Causeless, line 767: I'd prefer to either use ColorIndex as a property or even using  Get/SetColorIndex functions instead. We need to consider how to use this when palette is gone, and so being able to add a new unpaletted Color property would be better than having to break compatibility by changing the meaning of Color from a palette index to a "real" Color

Causeless (approved):

Causeless (changes requested):
As above comment about Color/ColorIndex

Source/Lua/LuaBindingsEntities.cpp
comradeshook: Good point, should be an easy fix; i'll just split it up into Get/SetColorIndex then

Causeless (approved):

PR #56, opened 2024-02-06
Teleporter fix

pawnishoovy:
had to use super global globals to make it work

also tiny Extinction fix

this ought to be it


Data/Base.rte/Scenes/Objects/Bunkers/BunkerSystems/Teleporters/TeleporterA.lua
```diff
@@ -1,13 +1,14 @@
 function Create(self)
 	-- Make sure the teleporter list exists.
 	if teleporterlista == nil then
-		teleporterlista = {};
+		_G["teleporterlista"] = {};
```
Causeless, line 4: oh jesus fuck

Data/Browncoats.rte/Devices/Shields/Eruptor/Eruptor.lua
```diff
@@ -0,0 +1,164 @@
...
+				self.SharpStanceOffset = self.bashActiveStanceOffset;
+			end
+		end
+	end
+
+end
+
+function SyncedUpdate(self)
+
+	if MovableMan:ValidMO(self.hitActor) then
```
Causeless, line 117: self.hitActor isn't being reset back to nil anywhere? This means another script that causes a synced update will cause this to happen again.

(note - I want to change this in future so that SyncedUpdate is sandboxed to each script, but it's still just a good idea for code cleanliness)

Causeless (changes requested):

Data/Browncoats.rte/Devices/Shields/Eruptor/Eruptor.lua
```diff
@@ -74,6 +77,7 @@ function ThreadedUpdate(self)
 				self.bashActive = false;
 				self.bashTimer:Reset();
 				self.bashCoolDown = true;
+				self.hitActor = nil;
```
Causeless, line 80: wouldn't it be safer to just set hitActor to nil at the very top of ThreadedUpdate?

Causeless (approved):

PR #61, opened 2024-02-10
Macos crosscompile

HeliumAnt:
- Adds cross compilation from linux to macos CI (this roughly halves compile time) 
- Update actions versions to most recent versions
- downgrade ccache action version to 1.12.11 removing node20 to workaround 2m timeout
- disable lto on linux ci except release bundles

First build will take ~20-30 min since it needs to build the compiler toolchain but subsequent builds will be normal. Toolchain is cached, however due to how actions cache access works, the cache will only be usable for all branches once it has run on the main branch (probably dev?)
https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#restrictions-for-accessing-a-cache

The cross compiler setup can also be updated to support arm64 (or all apple supported architectures)

(this took way longer than it had to because openssl has exactly one version which moved rmd160 to the legacy provider which of course is the one on ubuntu-latest (and accidental cache invalidation))


Causeless:
Awesome, thanks!

HeliumAnt:
still working on fixing macos packaging

Causeless (approved):

PR #99, opened 2024-02-18
Activities: Fix more player 1 assumptions, make broken code to avoid deploying on a player work.

MageKing17:
Signal Hunt was only checking to see if speedrun mode should be disabled if Player 1 had a brainbot, so if you used a different player, you could activate speedrun mode after deploying and spending all your gold, somewhat undercutting the point of speedrun mode. The check now scans over every player to look for brainbots, and additionally, `DoSpeedrunMode()` also iterates over all players looking for the rocket instead of assuming Player 1 (actually a hardcoded index 0) is active.

On an unrelated note, `DoSecret()` now calls `SwitchToActor()` on the brainbot after changing its team so that it's not left under AI control until you swap back to it. It also skips setting the AI mode and target for player-controlled actors, so your brainbot won't get an automatic goto order. However, if using a player other than player 1, the HUD breaks for some reason; I have been unable to track down the reason why, and it doesn't appear to be due to an assumption about which player executed the code.

Decision Day now scrolls each player's camera to their own brainbot instead of the first player's brainbot, which (as per usual) broke if player 1 wasn't present.

Harvester was applying fog of war to `self:GetTeamOfPlayer(Activity.PLAYER_1)` without bothering to check if Player 1 was even active; since the human team is always Team 1, and the team of an inactive player is Team 1 anyway, I just changed it to that.

Six activities had a block of code using nonexistant variables to check if there was only 1 player, and if so, make the AI's deployment unable to pick a spot directly above the player's head. I changed it to instead use existing variables, so the code should now function properly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - `LERP` Lua binding has been deprecated, and renamed to `Lerp`.
 
+- Six activities (Harvester, Keepie-Uppie, Massacre, One-Man Army, One-Man Army (Diggers Only), and Survival) now avoid AI deployments on top of the player.
+
 </details>
 
 <details><summary><b>Fixed</b></summary>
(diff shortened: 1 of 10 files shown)


Causeless:
Nice work!

Causeless (approved):

PR #106, opened 2024-02-19
Fix Constructor instantly cancelling the build cursor sometimes.

MageKing17:
If "Order Construction" was actively selected instead of letting it be selected by releasing the pie menu button, the script would detect the `PIE_MENU_ACTIVE` control and immediately cancel the controller. I added a variable to track whether or not `PIE_MENU_ACTIVE` is still held from the option being selected, and to ignore it if so.


Causeless:
Awesome! This is fixing a long time issue, the community are gonna love it

Causeless (approved):

PR #107, opened 2024-02-20
Fixed issue regarding Supportable and SupportOffsetWhileReloading

Alejandroibo:
If a gun had SupportOffsetWhileReloading == true and you modified it's Supportable value to be only true while reloading (and false when not), the hand wouldn't support the gun on the reload. 

The one handed penalty is still applied to the reload, since the gun reloads quicker than the hand reaches the gun. 
Nothing a few lines of code in lua to restrict the reload can't fix for modders that care about it

### DIFF
--- a/Source/Entities/AHuman.cpp
+++ b/Source/Entities/AHuman.cpp
@@ -2243,7 +2243,7 @@ void AHuman::PreControllerUpdate() {
 				} else if (heldDevice) {
 					if (HeldDevice* bgDevice = GetEquippedBGItem(); bgDevice && !heldDevice->IsOneHanded()) {
 						UnequipBGArm();
-					} else if (!bgDevice && !heldDevice->IsReloading() && heldDevice->IsSupportable()) {
+					} else if (!bgDevice && (!heldDevice->IsReloading() || (heldDevice->IsReloading() && heldDevice->GetUseSupportOffsetWhileReloading())) && heldDevice->IsSupportable()) {
 						m_pBGArm->SetHeldDeviceThisArmIsTryingToSupport(heldDevice);
 
 						if (!m_pBGArm->HasAnyHandTargets() && m_pBGArm->GetHandHasReachedCurrentTarget()) {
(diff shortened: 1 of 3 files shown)


Alejandroibo:
I've found another bug present with the Supportable value ,however I'm a bit stumped as how to fix it:
Let's say that I have a gun A with Supportable = 0. If you switch to it for the first time, your offhand will go the arm's IdleOffset (Intented behaviour)
If now I change to another gun B with Supportable = 1, and then switch back to gun A, now the offhand is stuck trying to reach the last position of gun B's supportOffset, instead of going back to the arm's IdleOffset like before.

Source/Entities/AHuman.cpp
```diff
@@ -2243,7 +2243,7 @@ void AHuman::PreControllerUpdate() {
 				} else if (heldDevice) {
 					if (HeldDevice* bgDevice = GetEquippedBGItem(); bgDevice && !heldDevice->IsOneHanded()) {
 						UnequipBGArm();
-					} else if (!bgDevice && !heldDevice->IsReloading() && heldDevice->IsSupportable()) {
+					} else if (!bgDevice && (!heldDevice->IsReloading() || (heldDevice->IsReloading() && heldDevice->GetUseSupportOffsetWhileReloading())) && heldDevice->IsSupportable()) {
```
MageKing17, line 2246: Checking `IsReloading()` twice is redundant.
```suggestion
					} else if (!bgDevice && (!heldDevice->IsReloading() || heldDevice->GetUseSupportOffsetWhileReloading()) && heldDevice->IsSupportable()) {
```

MageKing17 (changes requested):

MageKing17 (approved):

MageKing17:
As far as the second problem goes, it looks like the problem is that `m_HeldDeviceThisArmIsTryingToSupport` is never checked to see if that device is actually, you know... held. The conditional on line 2268 could easily have an `|| !m_pBGArm->GetHeldDeviceThisArmIsTryingToSupport()->IsBeingHeld()` added to it.

MageKing17 (dismissed):

Causeless (approved):

Causeless (approved):

PR #117, opened 2024-02-27
Missing codeSequence table

Bebomonky:
Fixes an issue with **codeSequence** not being initialized thus causing an issue in the Update function on **line 99**
This is because the **codeSequenceOrCodeType** argument in **SecretCodeEntry.Setup()** function accepts either a number or table
But since the table doesn't get assigned to **codeSequence** (as it doesn't exist at all), it results in it being a nil value


Causeless:
Can't you check that it's not nil instead?

Also does this issue ever occur in vanilla or only mods?

MageKing17:
There's no "instead", because without this assignment the code sequence would never be available to the `Update()` function at all. The problem can't occur in vanilla because vanilla only ever passes numerical arguments.

Causeless (approved):

PR #126, opened 2024-04-02
Activities: More player/screen mixup fixes, Decision Day typo.

MageKing17:
`CameraMan:GetScrollTarget()` and `CameraMan:SetScrollTarget()` both take a screen index but were often being passed player indices. I've hopefully caught and fixed all such instances. In addition, when `Activity:ScreenOfPlayer()` was being called multiple times in a row, I stored the result of a single call and replaced the multiple calls with that local variable.

Lastly, when I changed Decision Day to use each player's brain position instead of always using Player 1's, I accidentally wrote `pos` instead of `Pos`, so the first argument would've always been `nil`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,13 +35,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 </details>
 
-<details><summary><b>Removed</b></summary>
-
-</details>
-
 <details><summary><b>Fixed</b></summary>
 
 - Fixed an issue where palette index 255 was incorrectly showing as black.
+
+- Fixed instances of `CameraMan:GetScrollTarget()` and `CameraMan:SetScrollTarget()` supplying a player index instead of a screen index.
+
+- Fixed a bug in Decision Day that could cause an error when trying to set the camera's scroll target, in addition to the previous issue.
 
 </details>
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Fixed an issue where palette index 255 was incorrectly showing as black.
 
-- Fixed instances of `CameraMan:GetScrollTarget()` and `CameraMan:SetScrollTarget()` supplying a player index instead of a screen index.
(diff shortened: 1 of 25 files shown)


CHANGELOG.md
```diff
@@ -35,14 +35,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
 
-<details><summary><b>Removed</b></summary>
-
-</details>
-
 <details><summary><b>Fixed</b></summary>
 
 - Fixed an issue where palette index 255 was incorrectly showing as black.
 
+- Fixed instances of `CameraMan:GetScrollTarget()` and `CameraMan:SetScrollTarget()` supplying a player index instead of a screen index.
```
Causeless: Please try to write the symptom that this is fixing, not the cause without context - this is a bit meaningless to anyone playing the game

Source/Managers/CameraMan.cpp
```diff
@@ -64,10 +64,16 @@ void CameraMan::SetScroll(const Vector& center, int screenId) {
 }
 
 Vector CameraMan::GetScrollTarget(int screenId) const {
+	if (screenId < 0 || screenId >= c_MaxScreenCount)
```
Causeless: no bracketless ifs

Source/Managers/CameraMan.cpp
```diff
@@ -64,10 +64,16 @@ void CameraMan::SetScroll(const Vector& center, int screenId) {
...
+		// Would it be preferable to just set screenId to 0?
+		return Vector();
 	return g_NetworkClient.IsConnectedAndRegistered() ? g_NetworkClient.GetFrameTarget() : m_Screens[screenId].ScrollTarget;
 }
 
 void CameraMan::SetScrollTarget(const Vector& targetCenter, float speed, int screenId) {
+	if (screenId < 0 || screenId >= c_MaxScreenCount)
+		return;
```
Causeless, line 76: no bracketless ifs

Causeless (changes requested):

Source/Managers/CameraMan.cpp
MageKing17: I copied it from `GameActivity::DrawGUI()`, but my preference is for brackets anyway.

Causeless (approved):

PR #130, opened 2024-05-08
Add Slow-Mo Buy Menu global script

jacenhan:
Added a new global script to Base.rte that slows down the game to 1/10th speed when the Buy Menu is open. Uses the current method of slow-motion that manually changes DeltaTime and TimeScale, but it seems to work pretty well. AFAIK there is no way to pause the simulation completely from Lua as of right now (setting the TimeScale to 0 causes all sorts of errors).

This involved opening up the IsBuyGUIVisible function to be able to access it from Lua in ActivityMan.

### DIFF
--- a/Data/Base.rte/Scripts/Global/BuyMenuSlowMo.lua
+++ b/Data/Base.rte/Scripts/Global/BuyMenuSlowMo.lua
@@ -1,0 +1,39 @@
+function BuyMenuSlowMoScript:StartScript()
+	self.SlowMo = false;
+	self.Change = false;
+	self.originalTimeScale = TimerMan.TimeScale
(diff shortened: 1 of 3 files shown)


Causeless:
So in principle this is okay, but has some serious downsides when it comes to determinism and that the game doesn't work at different DTs altogether well. However, being a global script and not in the C++ (and being that we're not *really* deterministic yet anyways...), this should be fine.

pawnishoovy:
yep the old timescale + deltatime trick. as a global script it's fine, if you enable a weirder global script you implicitly agree to your game not being pixel-perfect anyway. only thing i'd add here is mentioning the new exposed function and the globalscript in the changelog.

Causeless (approved):

PR #132, opened 2024-05-27
Fog of War and Activity Settings Fixes

Bit-Crust:
Fog resolution increased dramatically, reveal by sight, scrambler rework, scanner rework, activity settings no longer ignored, out of place activity settings no longer enabled.


Data/Base.rte/Activities/SkirmishDefense.lua
```diff
@@ -268,35 +268,38 @@ function SkirmishDefense:UpdateActivity()
 
 			-- Add fog of war once the game is no longer in editing mode.
 			if self.addFogOfWar then
-				self.addFogOfWar = false;
+				local fogResolution = 1;
```
Causeless: Should these fogResolution defines perhaps be in SceneMan on the CPP side?

Data/Base.rte/Activities/SkirmishDefense.lua
```diff
@@ -268,35 +268,38 @@ function SkirmishDefense:UpdateActivity()
...
+							SceneMan:RevealUnseenBox(x - 10, 0, fogResolution + 20, altitude + 10, self:GetTeamOfPlayer(player));
+						end
 					end
 				end
 
-				for team = 0, Activity.MAXTEAMCOUNT - 1 do
+				for team = Activity.TEAM_1, Activity.MAXTEAMCOUNT - 1 do
 					if self:TeamActive(team) and self:TeamIsCPU(team) then
-						SceneMan:MakeAllUnseen(Vector(65, 65), team);
+						local fogResolution = 65;
```
Causeless: is this 65 intentional?

Source/Managers/SceneMan.h
```diff
@@ -510,6 +510,23 @@ namespace RTE {
...
+		/// @param ray The vector to trace along.
+		/// @param endPos A Vector that will be set to the position of where the sight ray was
+		/// terminated. If it reached the end, it will be set to the end of the ray.
+		/// @param size The minimum size of the box as it's cast.
+		/// @param strengthLimit The material strength limit where
+		/// @param skip For every pixel checked along the line, how many to skip between them
+		/// for optimization reasons. 0 = every pixel is checked.
+		/// @param reveal Whether the ray should reveal or restore unseen layer
+		/// @return Whether any unseen pixels were revealed as a result of this seeing.
+		bool CastUnseenBox(int team, const Vector& start, const Vector& ray, Vector& endPos, int size, int strengthLimit, int skip, bool reveal);
```
Causeless: Have you considered casting a circle instead of an AABB? Might give a prettier rounded revealed area at corners

Causeless:
Looks great! Few questions ad comments

Source/Managers/SceneMan.h
Bit-Crust: I did consider it back in the Lua tests, it actually looks really bad on flat walls. Circles have a really hard time filling in the depth of the wall, and show more gratingly that there are not a lot of casts. When I had it running, everyone who saw told me they liked how the corners looked before anyhow. A [Squircle](https://en.wikipedia.org/wiki/Squircle) might be a good middle ground, but it'd have to scale to any dimensions, so I don't know if that'd get costly performance wise.

Data/Base.rte/Activities/SkirmishDefense.lua
Bit-Crust: I don't know why but they were already in the game, so I left them, but there's no good reason for that aside from the player never seeing the enemy FoW, when I didn't leave anything similar in other activities. I've got another bugfix to push, so I'll standardize that while I go.

Data/Base.rte/Activities/SkirmishDefense.lua
Bit-Crust: I left it activity side because I think the whole activity's fog handling should be up to the activity. Same for unit deploying, honestly, Harvester was throwing 20 assertion fails (edit: in the release build) and assigning you units of the wrong faction because units were being deployed to Team II in C++ and then Lua tried to cover up. That's fixed now, but I think activity setup has a ways to go.

Causeless:
Awesome, thanks for the quick response to my questions - I'm happy with all of this. You mentioned you have some more commits coming, so let me know when this is ready and I'll merge (otherwise you can set this to draft until you're ready)

Causeless (dismissed):

Causeless:
Oh, please update changelog if you can, too!

Causeless (approved):

PR #134, opened 2024-06-03
allow/enable Bitwise operations library for scripts.

goodusername123:
Makes it so that scripts are allowed to access and use the Bitwise operations module (also known as BitOp) that is bundled with LuaJIT.

See here for info about the functions provided by this: https://bitop.luajit.org/api.html


Causeless (approved):

Causeless:
Nice one!

PR #140, opened 2024-09-01
Fix Mac build

HeliumAnt:
`>:(`

### DIFF
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -175,6 +175,8 @@ jobs:
         env:
           OSXCROSS_MACPORTS_MIRROR: "https://packages.macports.org"
         run: |
+          echo "::group::Installing pkg-config"
+          sudo apt install pkg-config
           echo "::group::Installing mac deps"
           omp install libsdl2 onetbb lz4 libpng minizip luajit flac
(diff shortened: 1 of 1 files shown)


.github/workflows/meson.yml
```diff
@@ -84,7 +84,7 @@ jobs:
           sudo pip install meson
 
       - name: ccache
-        uses: hendrikmuhs/ccache-action@v1.2.11 # Keep this at 1.2.11 until #181 or #182 is closed
+        uses: hendrikmuhs/ccache-action@v1 # Keep this at 1.2.11 until #181 or #182 is closed
```
Causeless: should we remove the comment?

.github/workflows/meson.yml
```diff
@@ -191,7 +193,7 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
 
       - name: ccache
-        uses: hendrikmuhs/ccache-action@v1.2.11 # Keep this at 1.2.11 until #181 or #182 is closed
+        uses: hendrikmuhs/ccache-action@v1 # Keep this at 1.2.11 until #181 or #182 is closed
```
Causeless: here too

.github/workflows/meson.yml
HeliumAnt: yes, both instances

Causeless (approved):

PR #141, opened 2024-09-12
Disable clang-format ci

HeliumAnt:


Causeless:
clang format was great but yeah, time to remove it from the build process... it's a bit too opinionated to keep now that the codebase is a lot cleaner with your contributions. approving this, thanks

Causeless (approved):

PR #145, opened 2024-10-06
Improved Navigation

Causeless:
Adds many enhancements that allow actors to locomote much more capably without using jetpacks.

* Improved high-speed limbpaths to work much more reliably
* Can now press control to crouch
* Can press shift to run


pawnishoovy (dismissed):
only thing worth mentioning is that the auto-stretching of the run path is a little arbitrary in its values but IMO it's completely unnecessary to make it a property. otherwise all this stuff is grand

pawnishoovy (approved):
i hereby approve my tiny revert

PR #146, opened 2024-10-11
Correct Dual Analog control mapping wizard.

Bit-Crust:
Lacking bitmaps and incorrect step count caused crashes in control setup wizard.


Causeless (approved):
Beautiful, nice catch

Issue #147, opened 2024-10-21
Parallelized call to CastSeeRays in MovableMan::Update tries to read null pointer


tjohnman:
**Describe the bug**
The parallelized loop that calls CastSeeRays on each of the m_Actors encountered a null pointer.
On inspection, none of the pointers in the deque are NULL. But due to it being parallelized, perhaps a NULL check should be done before trying to call the CastSeeRays method.

**To Reproduce**
This happened at the start of an attack in conquest mode without having first scanned the site.

**Expected behavior**
Game should not crash.

**Additional context**
```
Exception thrown at 0x00007FF759B73846 in Cortex Command.debug.release.exe: 0xC0000005: Access violation reading location 0x0000000000000000.
```
Call stack:
```
Cortex Command.debug.release.exe!RTE::MovableMan::Update::__l30::<lambda_2>::operator()(int start, int end) Line 1374	C++
```

OpenTools-run:
https://github.com/cortex-command-community/Cortex-Command-Community-Project/pull/148 already merged, please close

PR #148, opened 2024-10-21
Check for valid pointer before calling CastSeeRays method.

tjohnman:
Mitigates #147.


Causeless (approved):
Approving this, but adding some comments as there's a serious issue that will need fixed sooner or later

PR #149, opened 2024-10-29
Gibs and detached Attachables inherit rotational velocity from parent

comradeshook:
Through testing i found that gibbed MOSR's don't impart any of their AngularVel or offset-derived velocity from this AngularVel to their gibs. They do that now, and have this scaled by InheritsVel.

Additionally i found that detached attachables only inherited the plain velocity from the parent, similar to gibs. They now inherit rotational velocity in the same manner as gibs when detached.


Causeless (approved):
Nice work!

PR #150, opened 2024-11-02
GPU (SceneLayer) Drawing

HeliumAnt:
TODO:
- [x] Fix Windows build
  - [x] Add SDL_image to build
  - [x] Add raylib to build
  - [x] Rebuild allegro
- [x] Some sort of gpu memory management, otherwise this is a gigantic vram memory leak on scene change
- [x] Fix UnseenLayer not Drawing
- [x] Fix translucent UI elements
- [x] Documentation :)
- [x] ~~Split screen unseen layer at seams and screen edge~~ Scaled scenelayer upload
- [x] Space station is rotating the wrong way round
- [x] Glow strength
- [x] Rerun background removal
- ~~[ ] Primitives~~ (will do in followup PR)
---
NOTE: This VERY PROBABLY to DEFINITELY breaks online Multiplayer

# General Fixes
- Fixed allegro palettes being 6bit/color
- Retracted all allegro file interaction privileges, any image loading is now handled by SDL_image. (allegro likes to mess with colors, and does not load alpha values)
- Made MASK_COLOR_32 useful, by making it be 0 (black, alpha 0)

# GPU Drawing
- Added raylib for gl abstraction (does have some problems with state management, but is workable, may need to reimplement at some point)
- Infinite FPS
- Added StaticSceneLayer for non drawing layers (i.e. SLBackground, and TitleScreen nebula)
- Moved Main Menu Background to GPU
- Added Alpha channel to MainMenu Bitmaps
- SceneLayer drawing is pretty much entirely GPU now, any dynamic SceneLayers (everything other than SLBackground basically). 
- Big scenes will be drawn tiled, since they may otherwise not fit into vram (Refinery Assault uses ~800MB VRAM (technically not necessary, but this should make multistage FOW easier)
- Any DrawTrans is gpu drawn (except primitives, that's whole new can of fish)
- Made shaders loadable entities (rn still not using that fact particularly well, but oh well)


Causeless:
Very nice work! At this rate Release 8 CC will practically be a modern game with a modern architecture 💯  

I tried before the tiled drawing was implemented and there were some perf issues due to refinery assault's huge map, so now that that's done I'll try to test this again soon. Only other comment is that the clamp fix should be cherry-picked to another branch, as even though it's a fix it seems to regress crouching behaviour. I'll do that if you don't get to it by the time I do, so don't worry about it too much

Causeless (approved):

PR #151, opened 2024-11-14
Activity Scene Compatibility

Bit-Crust:
Activities need to somehow determine whether they are compatible with any given scene in the scenario menu. Until now this has been decided by two checks, one check for which areas are strictly required by the activity, determined by reading the script source file for every instance of `Scene::GetArea`, and another check where the Lua is allowed to arbitrarily determine which are compatible, communicated by the invalidation or failure to invalidate a Lua global.

The branch preserves both functions but changes their implementations. The strict areas are enumerated by any number of .ini properties "AddRequiredArea," each with the required area name specified, while the activity script method needs only to return true or false as to whether the given scene passes the test.

Internally, this is enabled by `LuaStateWrapper::RunScriptConditionalTestFunctionObject`, which does exactly the same thing as the function it was copied from, however it takes reference to a Boolean value as return parameter for the actual output of the conditional test. GetOptionalArea is also removed as well as all reference to it.

I think that's it, TL;DR please review the new function in LuaMan.cpp and tell me if there's anything odd about it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,9 +50,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - New `AEmitter` and `PEmitter` INI and Lua (R/W) property `PlayBurstSound` which denotes whether the BurstSound should play when appropriate. This should not be confused for a trigger - it's just a enable/disable toggle to avoid having to remove and add BurstSound altogether.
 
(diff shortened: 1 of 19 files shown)


Causeless (dismissed):
Beautiful! Very, very nice.

@pawnishoovy can you make sure that refinery assault gets a required area set up for this?

Causeless (approved):

PR #152, opened 2024-11-14
Default and locked team techs in Activities.

Bit-Crust:
Added Activity .ini properties TeamNTechSwitchEnabled, updated property TeamNTech to not safety check for installed mods, as the code actually can not know which mods are installed until after loading. Updated ComboBox code to not display the ComboButton while disabled, and display content at full width when disabled. Updated the ScenarioActivityConfigGUI to respect which faction is default selected and whether the team's faction is editable. Now an activity may specify default factions and lock the important ones.


Causeless (approved):
Nice work! This is some great functionality, especially for upcoming browncoat mission

Causeless:
@pawnishoovy can we investigate using this to force the enemy to be browncoats in refinery assault?

PR #153, opened 2024-11-14
Add ParticleCount to emissions, updates to InheritsVel

comradeshook:
- Added `ParticleCount` to Emissions which determines how many particles an emission fires out at once, defaults to 1 and has no effect on burst.
- Added `InheritsAngularVel` to both Gibs and Emissions, so as to allow things to inherit velocity without inheriting spin.
- Added `InheritsVelWhenDetached` and `InheritsAngularVelWhenDetached` to Attachables, functioning the same as the obvious parallels on Gibs and Emissions, but named for greater clarity.
- Updated `InheritsVel` on Emissions to account for the AngularVel of the parent
- Unclamped `InheritsVel` and its ilk, allowing modders to get a little silly
- Exposed all of these to Lua

OF NOTE: Changing properties of gibs and emissions in Lua doesn't work as of this writing.


Causeless:
Do we want to merge this now or will the issues with Gibs not being modifiable from Lua being sorted first? I'm not too fussed given how awkward it is here, but I heard some chat between yourself and @Bit-Crust there so I'm unsure if you're planning on adjusting that

Bit-Crust:
> Do we want to merge this now or will the issues with Gibs not being modifiable from Lua being sorted first? I'm not too fussed given how awkward it is here, but I heard some chat between yourself and @Bit-Crust there so I'm unsure if you're planning on adjusting that

We decided it ought to be a separate branch

Causeless (approved):

PR #154, opened 2024-11-14
Use local scope to hide proxies for raw lua functions.

Bit-Crust:
Functions OriginalDoFile, OriginalLoadFile, and OriginalRequire, which are what they sound like, removed from global scope by localizing them to "do end" blocks in source.


Causeless (approved):
First time I've even heard of do... end and I've been scripting Lua for a long time! Nice fix

PR #158, opened 2024-11-18
GPU Hotfix

HeliumAnt:
- Fix glow strength
- Initial z order support
- "Fix" scene editor (it'll look a bit odd in places, mainly cause i can't entirely fix the draw order for now, but for the most part it should be workable) proper fix needs full MO gpu draw support since the Editor GUIs all implement their own drawing stuff (currently terrain might draw in front of bunker pieces, and the pie menu punches holes in the placed tiles but thats all)


Causeless (approved):
Sounds fair overall, and next steps in description make sense too.

When we get GPU draw + z-order in for MOs too, would also make sense to investigate exposing that to Lua.

PR #160, opened 2024-11-20
Lua Writable Collections

Bit-Crust:
A number of properties which were exposed to Lua as iterators could not have their elements modified. When an iterator was obtained for Lua, it returned an iterator over a list of copies of elements from the original C++ collection, meaning that changes to those copies would be ineffective. Affected collections included MOSR Gibs, AEmitter or PEmitter Emissions, and MetaMan Players, unaffected were all collections of pointers. Corrected by changing all of those lists to pointers.


Causeless:
Oof, I can understand this but it feels really awkward doing it like this. It has performance implications, it messies the code, and it introduces a lot of manual memory management.

This is probably the safest approach, but at the same time it feels really awkward to have to do this everywhere we want to expose things to Lua...

Some thoughts:
1. We could instead have an intermediate lua binding that constructs a list of pointers to the Gib in the vector, instead of intrinisically actually allocating it seperately. This would avoid manual memory management and would also be more performant for the main C++ sim stuff as it'd avoid pointer chasing and potential cache misses there. However, if the vectors are modified, the elements within would become invalidated and bring the risk of Lua stamping over the wrong memory. To be fair this is already an issue that scripters are somewhat accustomed to dealing with though, as MO references in Lua aren't guaranteed to be stable for any longer than a frame.
2. We could do what we're doing here, but use unique ptrs instead of raw ptrs. Would be basically the same and still a bit awkward, but at least we get to avoid manual memory management.
3. We could have some entirely object that represents an entity id + index in the list and dereferences back onto the real object, while seeming just like a reference on the lua side. This would still be potentially risky but would be much less so

Interested in your thoughts

Causeless:
Think it's still worth getting this in, so merging

Causeless (approved):

Bit-Crust:
I disabled automerge for the moment, could've sworn I had an outgoing commit to push first, but false alarm. I can't turn it back on, however, presumably until the checks pass.

PR #163, opened 2024-12-06
Inventory Management

Bit-Crust:
Removes code to put away offhand automatically, and code to take out offhand automatically, both of which are typically aggravating in stressful situations where they may be triggered in error, such as trying to scurry in cramped halls while holding a shield.

Also adds code to ensure an offhand is discarded to the same side of the inventory as the onhand.

This is prelude to a minor inventory overhaul aimed at keeping order over time so that your inventory doesn't shuffle unless you're trying to shuffle it, but I don't know how exactly that'd work yet.


Causeless (dismissed):
Nice! Should make things a bit less annoying

Causeless (approved):

PR #164, opened 2024-12-07
Fix embarrassing SupportedGameVersion crash

pawnishoovy:
shuffles some code around so a missing SupportedGameVersion correctly warns the player (and lets him ignore it) instead of throwing a meaningless error message and crashing out.


Causeless (approved):
Nice catch

PR #165, opened 2024-12-07
Fix exit wounds not bursting and prevent multiple wound sounds per frame

comradeshook:
Fixes exit wounds not triggering their burst, prevents entry- and exit wound burst sounds from playing more than once per frame per MOSR, added optional isEntryWound and isExitWound bools to AddWound in Lua


Causeless (approved):
Beautiful, nice stuff!

PR #166, opened 2024-12-07
Zipped Save Files

Causeless:
Packages saves into a zip, reducing disk usage and simplifying save management.


Causeless:
Remaining work here is that packing the .pngs into the zip was tricky because of Allegro inexplicably missing functions to serialise/deserialise pngs from memory instead of from a file. This may not be an issue now with HeliumAnt's GPU layer draws, but I need to investigate again.

PR #167, opened 2024-12-07
Optimise Techion effects

comradeshook:
Ended up mostly being the Laser Rifle, which had _horrible_ performance due to spawning several hundreds of particles on every active frame. Additionally implemented a bit of optimisation tricks on the Dihelical Cannon and the disintegration effect for easy but relatively slight performance gains.


Causeless (approved):
Nice!

PR #168, opened 2024-12-09
Rays and hotkeys

pawnishoovy:
CastFindMORay now only finds exactly the specified MOID, not children

CastMORay now can also accept a table for ignored MOIDs, both Lua and C++

New CastAllMOsRay that returns an iterator of pointers to all MOs found along a ray (this one only accepts a table for ignored MOIDs)

New GetMOsAtPosition that returns an iterator of pointers to all MOs found at an exact position

Added all the new hotkeys to the config wizards and added default D-Pad bindings to controllers (this is particularly untested, I don't have a controller)

Fixed some inputs not having room in the binding menu, and also exposed all bindings for all control types


Causeless (dismissed):
Nice

Causeless (approved):

PR #169, opened 2024-12-09
Piemenu Behavior

Bit-Crust:
Fixes a couple problems with the way pie menus and aiming interacts between players, like opening the pie menu always resetting the M&KB player's aim and pie selection. Other commit potentially fixes an odd behavior which multiple had reported where a gamepad player's pie menu would suddenly fail to display, but since it could not be reproduced, I can also not test it, and haven't encountered it since making the branch. So, pull request.


Causeless (approved):
Awesome! This is a lognstanding issue we never quite managed to track down

PR #173, opened 2024-12-20
Macos Arm build

HeliumAnt:
- fix macos build
- add macos ARM CI


Causeless:
What extra work needs to be done here? Just testing etc? Also- can we distribute for ARM Linux too?

FWIW I was running native ARM when I was developing on Mac M1 and it worked perfectly, so I don't imagine we should see any issues here.

HeliumAnt:
yes just testing. The linux flatpak release already includes ARM builds, all that needs to be done there is add CI builds. I'll probably want to move the linux CI to target the Freedesktop SDK as well.

Source/Managers/WindowMan.cpp
```diff
@@ -101,8 +101,8 @@ void WindowMan::Initialize() {
 
 	SDL_GL_SetAttribute(SDL_GL_DOUBLEBUFFER, 1);
 	SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
-	SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 3);
-	SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);
+	SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 1);
```
Causeless, line 126: Are these changes still necessary/valid with the changes in b5b183e?

Causeless: Oh, that CL is older. No idea why it's merge conflicting, but eh

PR #174, opened 2024-12-20
Linux and MacOS backtrace

HeliumAnt:
Adds pretty printed backtrace and signal handler for non windows systems using [bombela/backward-cpp](https://github.com/bombela/backward-cpp)
(This could also do stacktrace on windows)


Causeless (changes requested):
Looks good, but can we do this for Windows too? I'd rather avoid having two stack trace libs

HeliumAnt:
Should pretty print backtrace on win as well now. (Pls test)

Causeless (approved):

PR #176, opened 2025-01-01
Fix tunnel slopes

HeliumAnt:
those looked bad enough that I had to learn some pixel art


Causeless:
Before merge I think I gotta update credits

Causeless (approved):

PR #182, opened 2025-01-20
Scripts Maintaining Insertion Order

Bit-Crust:
Previously, scripts were recorded with an unordered map from script name to enabled state, however this doesn't guarantee maintained order, causing script failure when behavior in the Create function (or elsewhere) of one script depends on the behavior of other scripts, which may be scheduled in certain circumstances to occur too late. An ordered map wouldn't work because it requires a comparison function, when the order needs to be determined by insertion order. It seems like it may as well be handled with a vector of script names, while maintaining the unordered map of script name to enabled state. However, each script name is recorded twice here (this is also what I have implemented, it works). It seems somehow less than ideal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -187,6 +187,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Fixed several issues with the way pie menus and aiming interacts between players, such as opening the pie menu always resetting the M&KB player's aim and pie selection, as well as another issue where the pie menu would fail to appear entirely for some players.
 
+- Fixed issue where scripts applied to `MovableObject`s could become disordered in certain circumstances.
+
 </details>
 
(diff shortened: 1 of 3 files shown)


pawnishoovy:
don't know enough c++ to comment on the technicalities but can confirm this keeps script order in-game by repeating steps in my mods that would previously break the order (and now they don't). @Causeless ?

HeliumAnt:
You could do a `std::vector<std::pair<std::string, bool>>`  to replace the unordered map, which would mean the initial code could stay mostly the same, pretty much, except for the insertion (`try_emplace` becomes `emplace_back`) and the search which would need to be a find_if with string compare lambda.

Might technically even be slightly faster as well, std::unordered_map performance is notoriously abysmal. Especially at these small entry numbers

Causeless (approved):

PR #186, opened 2025-02-09
Pvs studio fixes

janisozaur:
Some fixes spotted with PVS studio.

https://pvs-studio.com/en/order/open-source-license/


Source/Entities/MOSprite.cpp
```diff
@@ -269,8 +269,6 @@ bool MOSprite::HitTestAtPixel(int pixelX, int pixelY, bool validOnly) const {
 }
 
 void MOSprite::SetFrame(unsigned int newFrame) {
```
HeliumAnt, line 271: pretty sure this is lua accessible so I'd rather this accept `int` instead, to avoid underflow issues

Source/Entities/MOSParticle.cpp
```diff
@@ -126,12 +126,6 @@ void MOSParticle::Travel() {
 	m_Atom->ClearMOIDIgnoreList();
 
 	if (m_SpriteAnimMode == ONCOLLIDE) {
-		// Change angular velocity after collision.
-		if (hitCount >= 1) {
```
HeliumAnt, line 130: this also makes hitCount unused

HeliumAnt (changes requested):
Thanks!

Source/Entities/ACraft.cpp
zomgtehderpy, line 1: pretty sure this borks delivery in some cases

HeliumAnt: yup, those `break`s seem to be actually intentional. it's what we like to call Dan :tm: code (overly verbose code from the initial oss release that works mostly by luck)

HeliumAnt:
I think some of these really deserve a full refactor, since there's a few in here that are still mostly initial release code. (doesn't have to be here. I'll post up some issues)

Source/Entities/MOSParticle.cpp
Causeless: Maybe we should actually return hitcount from Atom travel instead?

Causeless:
Don't want to keep this laying around, but maybe a partial merge makes sense. We'll need to undo the changes to the DanCode™️that're breaking some stuff though.

PR #201, opened 2025-07-12
Add Windows window live resize

HeliumAnt:
Adds event watch for SDL_EVENT_WINDOW_EXPOSED to allow redraw while windows is locking the main thread during modal interactions (move, resize, etc.) 

~~(disabled on non windows since everywhere else those aren't blocking operations, and using the event watch leads to really bad flickering with vsync enabled)~~


Causeless:
This doesn't work- the screen just blacks out whenever resizing.

Is there no way to make screen resizes non-blocking on Windows?

HeliumAnt:
> Is there no way to make screen resizes non-blocking on Windows?

sortof, technically that is what I'm doing here, but apparently it's not working?
https://github.com/libsdl-org/SDL/issues/1059
https://stackoverflow.com/questions/3102074/win32-my-application-freezes-while-the-user-resizes-the-window

Causeless (approved):

PR #204, opened 2025-07-24
Pictorial Loadouts

Causeless:
Presets are now Loadouts, which have several changes:
* Now there is a visual representation of the loadout contents, with full text on hover, instead of just a huge text list
* Loadouts are now additive, and several can be mix-and-matched in your cart
* Now any loadout can be deleted, instead of it always choosing the last one in the list
* Loadouts can be drag-and-drop reordered much like the order cart list

image


MegaBaconator:
Amazing

PR #211, opened 2025-08-29
Miscellaneous Fixes

Causeless:
* Fixed issue where pie menu activation would become spotty at low framerates
* Fixed issue where game would freeze up after activity end


HeliumAnt (approved):
lgtm

PR #213, opened 2025-08-29
Fixed Seeray Crash

Causeless:
Fixes crash when script modifies actors that are seeing, and improves performance by giving a full frame of latency (with one frame delayed seeing)

Also fixes a minor issue where the buy menu GUI could ignore hover events if you clicked on certain elements to take focus away before closing and reopening the menu


HeliumAnt (approved):
:+1:

PR #228, opened 2025-11-03
LimbPath.cpp - fix debug line drawing

Architector4:
Debug draws in LimbPath.cpp don't take use the `targetPos` function argument and end up rendering wrong.

If only there was a compiler setting or something to find unused function arguments... :v

Before this patch:

https://github.com/user-attachments/assets/1853946c-d057-4a26-b0b1-d557635cd496

After this patch:

https://github.com/user-attachments/assets/7867325d-91e3-4701-8f8e-ef859734a566

The importance of this is probably "who cares", but at at least the squiggles aren't going to be all over the place with full debug mode now lmao


Causeless:
Huh. I could've sworn that this worked, at least at some point. Maybe with the changes to render direct-to-screen this fucked up. Nice catch, anyways!

PR #229, opened 2025-11-03
AHuman - rewrite EstimateJumpHeight

Architector4:
I got bored and set a scene to have zero gravity and soon found that the AIs caused this function to loop forever. I looked at it, was horrified by the code, yet felt it's simple enough to fix up in its entirety.

On quick testing, returns about the same results as the old code. Doesn't loop forever with no gravity though, and also accounts for burst fuel consumption a little better now lol


Causeless:
Very nice, much better!

PR #232, opened 2025-11-05
meson.build - enable debug compatible optimizations in debug modes

Architector4:
According to `man gcc`:

> Optimize debugging experience. -Og should be the optimization level of choice for the standard edit-compile-debug cycle, offering a reasonable level of optimization while maintaining fast compilation and a good debugging experience. It is a better choice than -O0 for producing debuggable code because some compiler passes that collect debug information are disabled at -O0.

TLDR: it's a no-brainer to enable for debug builds. I'm running the game in debug nowadays, and using this genuinely just makes it run way faster lol

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -68,8 +68,10 @@ if compiler.get_argument_syntax()== 'gcc' # used for gcc compatible compilers
     if debug_type == 'release'
       preprocessor_flags += ['-DRELEASE_BUILD'] # enable minimal debug features
     elif debug_type == 'minimal'
+      extra_args += ['-Og'] # debug-compatible optimizations
       preprocessor_flags += ['-DMIN_DEBUG_BUILD', '-DDEBUGMODE'] # enable some debug features
     elif debug_type == 'full'
+      extra_args += ['-Og'] # debug-compatible optimizations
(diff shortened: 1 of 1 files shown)


HeliumAnt:
Do not hardcode these in the build files, setup your build environment correctly instead using `meson configure --debug` or `debugoptimized` buildtype. Flatpak releases are also built with debug symbols

Architector4:
Huh. Never underestimate the complexity C++ codebase build systems I guess lmao

Architector4:
Oh wait, it gets even deeper: build type `debug` specifically adds `-O0`, killing optimizations, but build type `debugoptimized` specifically adds `-O2`, adding many optimizations, which in turn removes some debug info. To achieve the actually proper optimizations for debugging, one needs to use `--optimization=g` specifically.

why is meson like this lmao

HeliumAnt:
it's all very well documented. Using the build system to do these things means it's much more compatible between compilers. also though gcc says `-Og` is optimized for debugging, it doesn't actually work that well. in a lot of cases it still causes symbols to be missing or misplaced. `-O2` mostly isn't even that much worse for debugging (so long as the symbols are generated)

Architector4:
Eh, fair. I guess I was just not expecting there to be a different debug build type in Meson, or this setting to not be that good apparently lol

Issue #221, opened 2025-10-20
Some Lua scripts use OnDestroy which is never called by anything


Architector4:
**Describe the bug**
Some Lua scripts (Browncoats's Heatlance flamethrower and some Refinery breakables) define a function `OnDestroy`, which is not a function called by any core game code or script. However, `Destroy` is defined.

In case of Heatlance, this means that when it gets destroyed as it's being activated, the fire noise persists in the location it was at permanently for the rest of the game. It's quite a loud and annoying noise too.

I'm on the development branch, for reference.

**To Reproduce**
1. Go to `Data/Browncoats.rte/Devices/Weapons/Heatlance/Heatlance.lua`
2. Put `print("blah blah")` into the `OnDestroy` function
3. Spawn then utterly explode the Heatlance in the game
4. See no prints in the console.

Architector4:
This is easily fixable by replacing `function OnDestroy` to `function Destroy` in the 5 affected scripts. I'm too lazy to make a whole new git branch and pull request for this at this time, sorry lmao

The scripts (courtesy of `grep "function OnDestroy"`):

```
Cortex-Command-Community-Project/Data/Browncoats.rte/Scenes/Objects/Breakables/RefineryGenerator/RefineryGenerator.lua
Cortex-Command-Community-Project/Data/Browncoats.rte/Scenes/Objects/Breakables/RefineryDrill/RefineryDrill.lua
Cortex-Command-Community-Project/Data/Browncoats.rte/Scenes/Objects/Breakables/RefineryConsole/RefineryConsole.lua
Cortex-Command-Community-Project/Data/Browncoats.rte/Scenes/Objects/Breakables/RefineryCameraServer/RefineryCameraServer.lua
Cortex-Command-Community-Project/Data/Browncoats.rte/Devices/Weapons/Heatlance/Heatlance.lua
```

Causeless:
Will close as completed for now. I'll probably just add an alias for OnDestroy if we find it in any lua scripts. I think OnDestroy is a nicer name, actually...

PR #234, opened 2025-11-19
Fix some Lua scripts - OnDestroy -> Destroy

Architector4:
Fixes *current instances* of https://github.com/cortex-command-community/Cortex-Command-Community-Project/issues/221

But ideally, there should probably be some kind of instrumentation that creates warnings about Lua functions that are defined but not used by anything to prevent issues like this.


HeliumAnt (approved):
👍

PR #235, opened 2025-11-19
gitignore - also ignore *.sh

Architector4:
Shell files are basically the Linux equivalent of WIndows's *.bat files, which are already ignored. This is handy for me because I like using scripts for things lol

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,6 +74,7 @@ luac.out
 *.x86_64
 *.hex
 *.bat
+*.sh
 CortexCommand
 CortexCommand_debug
 CortexCommand.AppImage


HeliumAnt (changes requested):
Not sure about this. We have a few scripts that occasionally need updating, especially for CI and releases (i know it's not that big a deal to force update but it's easily missed). Explicitly should be disabled in `external/` mainly for sdl, which uses a bunch of scripts in its build process (updating sdl is already enough work as is, having unexpected ci errors from suddenly missing files doesn't help).

Architector4:
...Huh, did not expect there to be critical .sh files but not .bat files. My bad.

How about excluding just `/*.sh` for files in root? Shouldn't touch all that stuff I think.

PR #238, opened 2025-11-26
Assortment of compiler warning cleanups

Architector4:
I built the project with `-Dwarning_level=2` (`-Wall -Wextra`) with Clang and went over warnings in a few files.

This doesn't comprehensively clean up all of them, only a few, but also a few tiny logic bugs I noticed. Kind of mostly looking for input on this, because I imagine there might be disagreements to some of those.

For instance:
* Clang warns about code like `if (var = MightReturnNull()) …` because the intent might have been to use `==` instead, and advises to use double parentheses around the function as a way to explicitly denote the intent of assignment. I did that a bunch of times.
* Clang warns about dividing an integer then casting to float because the other way around incurs less precision loss. This primarily applies to code that computes the center of a bitmap/whatnot with integer coordinates and stores the result to a `Vector` (which uses floats). I switched those around too.
* I cleaned a bunch of unused variables, function arguments, and even imported but never used headers. For those I tried paying attention as to which commit has introduced them and if there was any reason they existed at that time, but they pretty much all appear to come from the initial commit unused already I think lol
* One spot in `SceneMan.cpp` runs a function that returns a `std::vector` and then does `std::move` on that, which had Clang warn `Moving a temporary object prevents copy elision` and suggest just removing it, so, yeah lol
* But what if a metagame lasts for more than 99 million rounds?? ...said a Clang warning about `GetRoundName` using a too small `numStr` buffer, so I expanded it lol
* There was also an *"definition of implicit copy constructor is deprecated because it has a user-provided copy assignment operator"* warning about `Vector`, so I gave it an explicit copy constructor. There was also a vice-versa of that warning for `Timer` so I gave it an explicit copy assignment operator lol

With this patch the game still compiles just fine. I played a few scenarios and a metagame and everything seems dandy.

But yeah, looking for input.

### DIFF
--- a/Source/Entities/LimbPath.h
+++ b/Source/Entities/LimbPath.h
@@ -375,7 +375,7 @@ namespace RTE {
 		// The number of starting segments, counting into the path from its beginning,
 		// that upon restart of this path will be tried in reverse order till one which
 		// yields a starting position that is clear of terrain is found.
-		int m_StartSegCount;
+		size_t m_StartSegCount;
 
 		// Array containing the actual 'waypoints' or segments for the path.
 		std::deque<Vector> m_Segments;
--- a/Source/Menus/MetagameGUI.cpp
+++ b/Source/Menus/MetagameGUI.cpp
@@ -157,7 +157,7 @@ void MetagameGUI::Clear() {
 	m_AnimMetaPlayer = Players::NoPlayer;
 	m_AnimDefenseTeam = Activity::NoTeam;
 	m_AnimActivityChange = false;
-	Scene* m_pAnimScene = 0;
+	m_pAnimScene = 0;
 	m_AnimRatio = 0;
 	m_AnimProgress = 0;
 	m_AnimTotalFunds = 0;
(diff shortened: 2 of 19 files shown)


Architector4:
oh what the heck

Architector4:
ok now the CI passes

don't ask me on ethics of serializing a `size_t` ok 👍

Source/Entities/LimbPath.h
```diff
@@ -375,7 +375,7 @@ namespace RTE {
 		// The number of starting segments, counting into the path from its beginning,
 		// that upon restart of this path will be tried in reverse order till one which
 		// yields a starting position that is clear of terrain is found.
-		int m_StartSegCount;
+		size_t m_StartSegCount;
```
HeliumAnt: Pretty sure these can and are set below 0 occasionally

Source/Menus/MetagameGUI.cpp
```diff
@@ -594,7 +594,7 @@ void MetagameGUI::MoveLocationsIntoTheScreen() {
 
 	// We need to calculate planet center manually because m_PlanetCenter reflects coords of moving planet
 	// which is outside the screen when this is called first time
-	Vector planetCenter = Vector(m_RootBoxMaxWidth / 2, g_WindowMan.GetResY() / 2);
```
HeliumAnt: See above

Causeless (approved):
Very nice!

Architector4:
...I take it as that everything I listed and did here is just fine? lol

HeliumAnt:
these are (almost) entirely semantic/style warnings, there's not much you can do wrong. (though technically the float divisions were actually better before since doing it the "correct" way causes subpixel misalignment, which will potentially cause issues at some point) 
at the same time imo fixing these is neither harmful nor particularly helpful either. `Wextra` is ok to use, although the codebase still isn't in a state where those warnings are particularly meaningful.

PR #250, opened 2025-12-23
Improve LimbPath.cpp - rewrite ReportProgress, etc

Architector4:
The rewrite checks all path segments from the current one to the last and finds the best one to continue from, instead of fixating on just the one. This helps in cases where feet might be yanked along the path by excessive horizontal acceleration, or when the player switches look direction while moving. Removes random stumbling too, mostly.

This doesn't perfectly fix everything ever, but it definitely makes it better I think. It's now more feasible to spam the jetpack and literally drag feet on the ground when the character's baggage is too heavy lol

Also fix Vector/Matrix division, making it the proper inverse of multiplication and applicable to a const Matrix. Figuring out why the math did not math definitely made me pull a hair or two lmao

Played a round of Bunker Breach with this patch and everything appears fine. Going over the scene wrap border break anything either.

---

Old behavior:

https://github.com/user-attachments/assets/c75b9f5c-fba3-4879-a5e7-80b7307e9b27

New behavior:

https://github.com/user-attachments/assets/a6cd142e-0fd4-4983-9e53-41c6df89ed99

### DIFF
--- a/Source/Entities/LimbPath.cpp
+++ b/Source/Entities/LimbPath.cpp
@@ -337,9 +337,9 @@ void LimbPath::ReportProgress(const Vector& limbPos) {
 		// This rest of the code will be working in local space, so convert input limb pos to that.
 		Vector limbPosLocal = ToLocalSpace(limbPos);
 
-		//
+
 		// Iterate over all segments and find one whose target is closest to the limb position.
-		//
+
 
 		// Segment positions are accumulative, so keep an accumulator.
 		Vector currentSegmentStartPos = GetCurrentSegStartLocal(); // Will be needed later.
@@ -350,12 +350,27 @@ void LimbPath::ReportProgress(const Vector& limbPos) {
 		std::deque<Vector>::iterator closestSegment = m_CurrentSegment;
 
 		for (std::deque<Vector>::iterator itr = m_CurrentSegment; itr != m_Segments.end(); ++itr) {
-			if (itr != m_CurrentSegment) {
+
+			// We want to find a closest segment to work off of, but we don't want to
+			// snap from collision-enabled segments to collision-disabled segments,
+			// because doing so tends to produce erratic foot behavior.
+
+			// If the current segment of the limbpath is collision-enabled...
+			if (!FootCollisionsShouldBeDisabled()) {
+				// If the currently looked at segment is collision-disabled...
 				if (m_FootCollisionsDisabledSegment >= 0 &&
 						m_Segments.size() - (itr - m_Segments.begin()) <= m_FootCollisionsDisabledSegment) {
-					// We've already picked a segment (at least the current one),
-					// and the remaining ones are ones with collisions disabled.
-					// Ignore these.
+
+					// ...Then break.
+
+					// Note: if the first of the above two checks has passed,
+					// this means that the current segment is collision-enabled.
+					// And, since this iterator starts with it, this means that
+					// *at least* the current segment was picked as closest already.
+					//
+					// In other words, if this break was hit, then closest segment vars have
+					// been properly initialized already. Therefore, it's safe to break.
+
 					break;
 				}
 			}
@@ -369,16 +384,13 @@ void LimbPath::ReportProgress(const Vector& limbPos) {
 				closestSegmentStartPos = thisSegmentStartPos;
 				closestSegmentTargetDistanceSqr = thisSegmentDistanceSqr;
 				closestSegment = itr;
-			} else {
-				// This one is *farther* than the last one.
-				// Assuming next segments will be only farther and farther, just break now.
-				break;
 			}
 		}
 
-		// We will want to compute progress to whatever the new segment is.
-		float distanceToCurrentSegmentTargetSqr;
 
+		// Branches below will determine the new current segment and write the distance to it here.
+		// We need this distance to compute progress towards it, whatever it ends up being.
+		float distanceToCurrentSegmentTargetSqr;
 
 		if (closestSegmentTargetDistanceSqr < m_SegmentEndedThreshold * m_SegmentEndedThreshold) {
 			// We're sufficiently close to this segment's target to go on.
--- a/Source/Entities/LimbPath.cpp
+++ b/Source/Entities/LimbPath.cpp
@@ -172,6 +172,23 @@ Vector LimbPath::RotatePoint(const Vector& point) const {
 	return (((point - offset) * m_Rotation) + offset) + m_PositionOffset;
 }
 
+Vector LimbPath::InverseRotatePoint(const Vector& point) const {
+	Vector offset = (m_RotationOffset).GetXFlipped(m_HFlipped);
+	return (((point - m_PositionOffset) - offset) / m_Rotation) + offset;
+}
+
+Vector LimbPath::ToLocalSpace(const Vector& position) const {
+	// The position might be on one side of a border of a wrapping scene while the joint is on another.
+	// Account for that.
+	Vector posWrapped = m_JointPos + g_SceneMan.ShortestDistance(m_JointPos, position);
+
+	return InverseRotatePoint(posWrapped - m_JointPos) / GetTotalScaleMultiplier();
+}
+
+Vector LimbPath::ToWorldSpace(const Vector& position) const {
+	return m_JointPos + (RotatePoint(position * GetTotalScaleMultiplier()));
+}
+
 int LimbPath::Save(Writer& writer) const {
 	Entity::Save(writer);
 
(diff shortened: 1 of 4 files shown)


Source/System/Matrix.h
```diff
@@ -246,7 +247,10 @@ namespace RTE {
 		/// @param lhs A Vector reference as the left hand side operand.
 		/// @param rhs A Matrix reference as the right hand side operand.
 		/// @return A reference to the resulting Vector.
-		friend Vector operator/(const Vector& lhs, Matrix& rhs) { return rhs / lhs; }
+		friend Vector operator/(const Vector& lhs, const Matrix& rhs) {
+			Matrix m(rhs);
+			return m / lhs;
+		}
```
HeliumAnt, line 271: you can just mark the `Matrix::operator/(const Vector&)` const to avoid the copy.

Source/Entities/LimbPath.cpp
```diff
@@ -292,29 +310,119 @@ void LimbPath::ReportProgress(const Vector& limbPos) {
...
+		if (closestSegmentTargetDistanceSqr < m_SegmentEndedThreshold * m_SegmentEndedThreshold) {
+			// We're sufficiently close to this segment's target to go on.
+			// Either declare this path ended, or continue from the next segment.
+
+			if (closestSegment + 1 == m_Segments.end()) {
+				// Closest segment is the last segment and we are at its target. Declare done.
+				m_Ended = true;
+				m_CurrentSegment = closestSegment;
+
+				distanceToCurrentSegmentTarget = std::sqrt(closestSegmentTargetDistanceSqr);
```
HeliumAnt: sqrt is pretty expensive. since you're mostly just comparing distances, use the square magnitudes instead.

Source/Entities/LimbPath.cpp
```diff
@@ -292,29 +310,119 @@ void LimbPath::ReportProgress(const Vector& limbPos) {
...
+
+		// Now compute a normalized progress measure towards the current segment.
+
+		float currentSegmentMagnitude = m_CurrentSegment->GetMagnitude();
+
+		if (distanceToCurrentSegmentTarget > currentSegmentMagnitude)
+			// We're too far away from this target.
+			m_SegProgress = 0.0;
+		else
+			m_SegProgress = (1.0F - (distanceToCurrentSegmentTarget / currentSegmentMagnitude));
```
HeliumAnt: you can just sqrt once here instead.

Source/Entities/LimbPath.h
```diff
@@ -449,6 +451,26 @@ namespace RTE {
...
+		/// @returns Local space position
+		Vector ToLocalSpace(const Vector& position) const;
+
+		/// Converts an input position, in local space for this LimbPath, into scene/world space,
+		/// taking into account rotation and scaling.
+		/// @param position Local space position
+		/// @returns World space position
+		Vector ToWorldSpace(const Vector& position) const;
+
+		// Gets the current segment's starting position (i.e. last segment's target) in local space.
```
HeliumAnt: Doccomments use `///´

Source/Entities/LimbPath.h
```diff
@@ -383,7 +383,9 @@ namespace RTE {
 		// The iterator to the segment of the path that the limb ended up on the end of
 		std::deque<Vector>::iterator m_CurrentSegment;
 
-		int m_FootCollisionsDisabledSegment; //!< The segment after which foot collisions will be disabled for this limbpath, if it's for legs.
+		// Count of segments at the end of the segments list for which foot collisions should be disabled
+		// for this limbpath, if it's for legs.
```
HeliumAnt: `///` for multiline doccomments (always on function documentation) or `//!<` for short member documentation.

HeliumAnt: check git-blame. anything marked as initial commit should not be used for reference.

HeliumAnt (changes requested):
Awesome! Just some minor performance nitpicks and comment syntax.

Source/System/Matrix.h
Architector4: Can't do; the division operator, same as multiplication operator, potentially calls `Matrix::UpdateElements()` which needs to modify it.

I wrote this to be symmetric to `operator*` implementation just above this. It might be clearer to pass by-value instead of passing by const ref then copying here, but I'm unsure as to how that would impact performance of this function call lol

Source/Entities/LimbPath.cpp
Architector4: Oh yeah, true. Could be shaved off here until when the non-squared distance matters.

Source/Entities/LimbPath.h
Architector4: oops

Source/System/Matrix.h
HeliumAnt: Ah lol. This is essentially pass by value with extra steps. Value is probably clearer as to what it's doing (probably the same in terms of perf, it's doing the copy and update either way)

Source/Entities/LimbPath.h
HeliumAnt: Thanks for cleaning up the rest as well :)

Source/System/Matrix.h
Architector4: Eh, perhaps. I feel like the best approach would be to let the function take a mut ref and modify the original matrix and just let the code using it do a copy explicitly by itself if it really needs to. No extraneous copying, plus it would update elements of the original matrix properly.

I might do that but maybe another time lol

Causeless:
Sorry, GitHub mobile is shit and won't let me comment on specific lines, says an unknown error occurred.

So two thoughts:
1. I would rather we search everything for the closest. Walkpaths are not big, and they are also loops. It is very common that we may move further away from the outbound foot movement while the inbound foot movement is still closer.
2. Add a comment explaining the copy in matrix division. It's unclear why it's needed, and that code in isolation screams to me "this is dumb, why are we doing that?". Obviously it's required, though, so we should add a comment explaining that.

Architector4:
> Sorry, GitHub mobile is shit and won't let me comment on specific lines, says an unknown error occurred.

Happens to me too sometimes, especially when loading the page from an email for some reason; the fix seems to be to reload the page

> * I would rather we search everything for the closest. Walkpaths are not big, and they are also loops. It is very common that we may move further away from the outbound foot movement while the inbound foot movement is still closer.

Just to be clear, by "outbound" and "inbound" to do you mean collision-disabled and collision-enabled segments? I think I made the code not check the former because I figured that landing slightly too hard or carrying too much weight would periodically cause the code to focus both legs to those collision-disabled segments, promptly causing a torso landing. I should experiment I guess.

> 2\. Add a comment explaining the copy in matrix division. It's unclear why it's needed, and that code in isolation screams to me "this is dumb, why are we doing that?". Obviously it's required, though, so we should add a comment explaining that.

Yeah, I think I want to investigate if I can just avoid doing the dumb thing altogether. I wonder if I could make both by-mut-ref and by-const-ref versions of the operators lol

Causeless:
> > Sorry, GitHub mobile is shit and won't let me comment on specific lines, says an unknown error occurred.
> 
> 
> 
> Happens to me too sometimes, especially when loading the page from an email for some reason; the fix seems to be to reload the page
> 
> 
> 
> > * I would rather we search everything for the closest. Walkpaths are not big, and they are also loops. It is very common that we may move further away from the outbound foot movement while the inbound foot movement is still closer.
> 
> 
> 
> Just to be clear, by "outbound" and "inbound" to do you mean collision-disabled and collision-enabled segments? I think I made the code not check the former because I figured that landing slightly too hard or carrying too much weight would periodically cause the code to focus both legs to those collision-disabled segments, promptly causing a torso landing. I should experiment I guess.
> 
> 
> 
> > 2\. Add a comment explaining the copy in matrix division. It's unclear why it's needed, and that code in isolation screams to me "this is dumb, why are we doing that?". Obviously it's required, though, so we should add a comment explaining that.
> 
> 
> 
> Yeah, I think I want to investigate if I can just avoid doing the dumb thing altogether. I wonder if I could make both by-mut-ref and by-const-ref versions of the operators lol

I was originally talking about the early-out check in the code finding the closest segment. Once the next segment is further away than the current one, we immediately exit out and claim we've found the closest. 

This is very unreliable- depends on the order of the walkpath (for some given walkpaths, it could always return the furthest point!), but even for typical ones that involve a loop we could find a bad point 50% of the time.

With that said, we could experiment to find both collision enabled and disabled segments. Given we'd expect the walkpaths to be 180 degrees out of phase for both legs, though, in theory it would be best to actively force one leg to the collision-enabled-segments and the other to the disabled segments, at least when actively walking.

If the situation is that both feet are genuinely misplaced, maybe thats where we'd expect to see a genuine trip (and maybe we could trigger prone or something to make it feel more physically proper). But perhaps that's too aggressive here.

Architector4:
> I was originally talking about the early-out check in the code finding the closest segment. Once the next segment is further away than the current one, we immediately exit out and claim we've found the closest.

Ooh, yeah. I think I was just being too overly concerned with performance while making an assumption that seemed true from my observation of brain robot's walk path. There's not really that many segments and it could probably be allowed to iterate through all just fine lol

> With that said, we could experiment to find both collision enabled and disabled segments. Given we'd expect the walkpaths to be 180 degrees out of phase for both legs, though, in theory it would be best to actively force one leg to the collision-enabled-segments and the other to the disabled segments, at least when actively walking.

To note, I wrote the code to always pick *some* segment other than the current target to look at, and code that advances to the next segment if the target threshold is reached, together with that, basically guarantees that the non-colliding segments will all be went through anyway.

Hmm, I think a fair solution might be to skip collision-disabled segments iff the current segment is collision-enabled?

Causeless:
Yep, sounds reasonable. I think it ultimately depends on how it looks in-game, but in theory that sounds good.

Architector4:
Just did some checks.

The early break on any farther segment is indeed not needed, removing it doesn't seem to change much.

Just removing the part that skips collision-disabled segments is a bad idea; doing so causes some funky shuffling:

https://github.com/user-attachments/assets/e623666b-0ef4-4358-b33a-a9c10f15252c

> to skip collision-disabled segments iff the current segment is collision-enabled?

Put that into the code and it doesn't seem to change much at all. I guess it's a fair change to do anyway since it sounds like something that makes sense? lol

Causeless:
GitHub mobile won't let me approve this because Microsoft is investing $80 billion into glorified chatbots instead of maintaining their actually useful software, so I'm manually merging this in.

Thanks for the changes!

PR #254, opened 2025-12-30
MOPixel::Draw - don't skip drawing on non-drawn sim updates

Architector4:
MOPixels can be drawn onto the scene terrain in cases like digger particles settling. With this skip, digging with a `TimeScale` value more than 1 can cause those to skip rendering, accumulating ghost terrain that exists in material layer but not in color layer. Similar thing can happen with Constructor digging.

Without this fix (alternate terrain mode shown with RAlt+M): 

https://github.com/user-attachments/assets/15957168-346b-458b-9625-a9118b7b078a

With this fix: 

https://github.com/user-attachments/assets/817e7b82-14b1-41a6-aeff-4b20e5c5996a

I've also considered adding an extra check if this is drawing to scene terrain or not, but then I figured that this won't be reliable because what if we'll have multiple terrains in some weird future, or some other case where the target bitmap is persistent but isn't a scene terrain.

### DIFF
--- a/Source/Entities/MOPixel.cpp
+++ b/Source/Entities/MOPixel.cpp
@@ -222,10 +222,8 @@ void MOPixel::Update() {
 }
 
 void MOPixel::Draw(BITMAP* targetBitmap, const Vector& targetPos, DrawMode mode, bool onlyPhysical) const {
-	// Don't draw color if this isn't a drawing frame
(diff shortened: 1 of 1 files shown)


Causeless (approved):
At some point this drawing weirdness was going to be removed entirely when we get interpolated-render in, but this is a good change for now regardless. Approved, thanks!

PR #255, opened 2026-01-01
Fix save palette

HeliumAnt:


Source/System/ContentFile.cpp
```diff
@@ -63,7 +63,7 @@ void ContentFile::FreeAllLoaded() {
 int ContentFile::ReadProperty(const std::string_view& propName, Reader& reader) {
 	StartPropertyList(return Serializable::ReadProperty(propName, reader));
 
-	MatchForwards("FilePath") 
+	MatchForwards("FilePath");
```
Causeless: Unintentional? Pretty sure this breaks things

Causeless (changes requested):

Causeless (approved):

PR #259, opened 2026-01-08
Meson Configuration

Bit-Crust:
Meson configurations should be generally usable now. Removed conditional builds from all Meson build paths, removed expectations of differentiated executables from Meson tasks, renamed all tasks to accurately read either "Launch" or "Configure & Launch" depending on specifics, and created parity. Now the project can be either built and run or configured, built, and run, for both Windows and Unix configurations.

I have neglected to gut the MSBuild stuff because I dunno exactly how to do that besides just cutting it from the `tasks.json` and `launch.json` files. That can be a separate PR, I think.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -13,6 +13,21 @@
       "request": "launch",
       "program": "${workspaceFolder}/${config:mesonbuild.buildFolder}/CortexCommand",
       "cwd": "${workspaceFolder}",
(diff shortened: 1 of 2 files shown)


Bit-Crust:
I suppose gutting MSBuild also includes removing it from continuous integration and such.

HeliumAnt (approved):
Nice! You can leave msbuild as is, that's working fine.

Causeless (approved):

PR #266, opened 2026-02-04
Weapon & Activity Script Fixes

jacenhan:
Actually just two updates to fix things that bothered me when I was checking out the nightly build.

1. Bunker Breach now correctly spawns in door actors. More accurately, it no longer _deletes_ door actors, which it looks like it started doing because the actors themselves don't have the "Bunker Systems" group anymore. Not sure when that happened or how/if it should be fixed, but it looks like this is the only place that that check was being made, so I just added an IsADoor check and it works now.
2. Auto-targeting weapons (ML-02 Flash, Nucleo Swarm, Coalition Missile Launcher, Dropship turrets) no longer target actors with GetsHitByMOs set to false. This mainly affects mods that use "invisible" actors, like my Endless War/Breach mods, so this is a bit of a selfish one.


Causeless (approved):

Causeless:
nice!

PR #270, opened 2026-02-10
Separate thread module loading

getcetc:
The module loading work is now done in a worker thread.

This speeds up loading progress printout, if enabled. The is done by the worker thread only pushing the printout strings, while the main thread consumes and draws them out.

By consuming SDL events on the main thread, this:
- Makes the window not hang and go white after a few seconds during game launch. This popped an app-frozen error before.
- Allows for premature exiting via Alt+F4 or pressing the close button on the window.

This also sets up for future module loading multithreading work.
Asserts from worker thread work OK, but caution is to be taken on further MT work. Some scaffolding has been built for it.

Also implemented is a spinlock watchdog thread, asserting with useful info.


Source/Managers/PresetMan.h
```diff
@@ -7,12 +7,14 @@
 /// Inclusions of header files
 #include "Entity.h"
 #include "Singleton.h"
+#include "shader.h"
```
HeliumAnt, line 10: forward declare `class Shader` instead. And `#include "Shader.h"` in cpp

Source/Managers/PresetMan.cpp
```diff
@@ -350,6 +399,9 @@ const Entity* PresetMan::GetEntityPreset(Reader& reader) {
 	// Load class name and then preset instance
 	reader >> ClassName;
 	pClass = Entity::ClassInfo::GetClass(ClassName);
+	if (ClassName == "Shader") {
+		int a = 5;
```
HeliumAnt, line 403: ???? (though if you really _need_ this empty check to happen use `(void);` instead)

Source/Managers/PresetMan.cpp
```diff
@@ -108,72 +105,124 @@ bool PresetMan::LoadDataModule(const std::string& moduleName, bool official, boo
...
+	
+	// Main thread - we drain display queue and poll SDL events
+	//
+	// We do this so the window does not freeze, and to handle
+	// the close button on the window being hit / Alt+F4 pressed
+	{
+		while (1) {
+			mainThreadHeartbeat++;
+			PollSDLEventsCallback();
+			if (true) {
```
HeliumAnt, line 156: no need for the `if (true)`. you can just make a code block, that's still local scope.

Source/Managers/PresetMan.cpp
```diff
@@ -108,72 +105,124 @@ bool PresetMan::LoadDataModule(const std::string& moduleName, bool official, boo
...
+						break;
+					}
+
+					if (!m_ProgressDisplayDeque.empty()) {
+						entry = std::move(m_ProgressDisplayDeque.front());
+						m_ProgressDisplayDeque.pop_front();
+					} else if (spinlockDetected) {
+						spinlockDetected = false;
+						SpinlockAssert(toDoProgressPrintOut, loadingDone);
+						RTEAssert(false, to_string(mainThreadHeartbeat))
```
HeliumAnt, line 191: Btw we have `RTEAbort` for this usecase

HeliumAnt (changes requested):
Cool!
also linux/macos fixes needed

Source/Managers/PresetMan.cpp
getcetc: A very non-production PR from me, sorry! This and other things i intended not to have left in. Will fix up, thank you for great feedback!

Source/Managers/PresetMan.cpp
getcetc: I learn more things 🙂
Will use!

Source/Managers/PresetMan.cpp
getcetc: Ahh this was to set a breakpoint on it 😅

Source/Managers/PresetMan.h
getcetc: Good catch!

PR #272, opened 2026-03-04
Extra BrowncoatLight Hat

MrMikeX:
New browncoat light alternative hat addition


MrMikeX:
still not properly implemented, files need to be renamed accordingly to function with FrameCount, gib sprites need to be put in gibs

PR #282, opened 2026-06-12
Vendored nlohmann/json

Madreag:
Vendors the single-header nlohmann/json under `external/sources/`. Nothing uses it yet — it's the report format for the determinism test tooling later in this series.

nlohmann/json 3.12.0 (MIT, license included) is just the one `json.hpp` header, include-path only, nothing compiled. Wired into both build systems: a meson subproject and the include path in `RTEA.vcxproj`/`.filters`.

The second commit catches the changelog up — entries for #279, #280 and this vendor.

Built clean and boots with the full mod stack on Windows (MSBuild), Linux (meson) and macOS.

This is the first of the determinism series, split into small PRs to keep review easy. Next are the cross-platform stability fixes and the LuaJIT determinism build flags; it all builds up to a same-seed deterministic sim and the test harness that proves it — the groundwork for controller-synced multiplayer.


Causeless:
Bit odd to have a PR specifically for adding an unused library. I think it makes more sense to have it functionally have a use before we merge this in.

Madreag:
nothing uses it til the cccp-ctl determinism harness PR, so i'll just fold it into that one and close this out.

PR #283, opened 2026-06-20
Cross-platform stability fixes and FP-contract standardization

Madreag:
These are cross platform stability fixes that came up while bringing the determinism work over to Mac and Linux. Most are small correctness issues that Windows happened to hide (its heap fill pattern and thread scheduling masked a few), so they only showed up once the same code ran on the other two platforms. see commit messages.

* `AtomGroup` collision iteration now runs in MOID order (the two thread local `hitMOAtoms` maps go from `unordered_map` to `map`; `MOIgnoreMap` is left alone since it's lookup only and never iterated). Plus a guard against a 0/0 step ratio when every atom in a segment rasterizes to zero steps, which left a dead NaN that still tripped `FE_INVALID`.
* `HumanBehaviors` weapon and tool pickup sort gets a `deviceId` tie break, since the async pickup callbacks fire in scheduler dependent order and equal scores need a stable total order.
* FP contract pinned across toolchains: `-ffp-contract=off` on GCC/Clang and `FloatingPointModel=Precise` on every MSVC config, so the compiler can't fold FMA or reassociate floats differently per platform. A follow on fixes a Meson bug where the release branch reassigned `extra_args` and silently dropped those flags on every release TU (checked in `compile_commands.json`: 203 TUs now, was 0).
* `LuaAdapters` async callback id counter is now a `std::atomic<int>`, closing a race under parallel AI.
* Two uninitialized reads off the pooled allocator, both returning prior occupant bytes so the value tracked allocation order: `Atom`'s integer step state, and `MovableMan::m_SimUpdateFrameNumber`. Caught with Valgrind/ASan.
* An `AudioMan` use after free: an MO gibbed mid tick frees its `SoundContainer`, but its still playing FMOD channels keep the freed address in `userData`. The channels now get disowned on destroy (they keep playing) and the readers skip a container that's gone.
* A few Mac only fixes: message boxes raised from a worker thread now log to stderr instead of popping a dialog (a cross thread Cocoa alert deadlocks), an explicit `file_time_type` to `system_clock` conversion in the save menu, and the libc++ removed-API macros that luabind 0.7.1 and boost 1.75 still need to build on libc++ 19+.

* a one line guard in Tracy (`TracyProfiler.cpp`) that clamps `broadcastMsg.activeTime` to 0 when the timestamp predates the profiler epoch, otherwise it trips Tracy's own assert on startup.

Built and ran clean on Windows, Linux, and macOS arm64. A few of these (the iteration order, FP contract, and uninitialized reads) are prerequisites for the determinism series this is part of, but they all stand on their own as fixes.

### DIFF
--- a/Source/Managers/AudioMan.cpp
+++ b/Source/Managers/AudioMan.cpp
@@ -603,6 +603,19 @@ bool AudioMan::StopSoundContainerPlayingChannels(SoundContainer* soundContainer,
 	return result == FMOD_OK;
 }
 
+void AudioMan::DisownSoundContainerPlayingChannels(const SoundContainer* soundContainer) {
+	if (!m_AudioEnabled || !soundContainer || !soundContainer->IsBeingPlayed()) {
+		return;
+	}
+	// Leave the channels playing, but null their back-reference so the positional/ended passes don't read the freed container.
+	FMOD::Channel* soundChannel;
+	for (int channelIndex: *soundContainer->GetPlayingChannels()) {
+		if (m_AudioSystem->getChannel(channelIndex, &soundChannel) == FMOD_OK) {
+			soundChannel->setUserData(nullptr);
+		}
+	}
+}
+
 void AudioMan::FadeOutSoundContainerPlayingChannels(SoundContainer* soundContainer, int fadeOutTime) {
 	if (!m_AudioEnabled || !soundContainer || !soundContainer->IsBeingPlayed()) {
 		return;
@@ -673,7 +686,7 @@ void AudioMan::Update3DEffectsForSFXChannels() {
 				float doubleMinimumDistanceForPanning = m_MinimumDistanceForPanning * 2.0F;
 				void* userData;
 				result = result == FMOD_OK ? soundChannel->getUserData(&userData) : result;
-				if (result == FMOD_OK) {
+				if (result == FMOD_OK && userData != nullptr) {
 					const SoundContainer* soundContainer = static_cast<SoundContainer*>(userData);
 					if (sqrDistanceToPlayer < (m_MinimumDistanceForPanning * m_MinimumDistanceForPanning) || soundContainer->GetCustomPanValue() != 0.0f) {
 						result = soundChannel->set3DLevel(0);
@@ -702,6 +715,9 @@ FMOD_RESULT AudioMan::UpdatePositionalEffectsForSoundChannel(FMOD::Channel* soun
 	}
 
 	const SoundContainer* channelSoundContainer = static_cast<SoundContainer*>(userData);
+	if (channelSoundContainer == nullptr) {
+		return FMOD_OK; // The owning SoundContainer was destroyed; leave the channel playing where it is.
+	}
 
 	bool sceneWraps = g_SceneMan.SceneWrapsX();
 
@@ -802,7 +818,7 @@ FMOD_RESULT F_CALLBACK AudioMan::SoundChannelEndedCallback(FMOD_CHANNELCONTROL* 
 		result = (result == FMOD_OK) ? channel->getUserData(&userData) : result;
 		if (result == FMOD_OK) {
 			SoundContainer* channelSoundContainer = static_cast<SoundContainer*>(userData);
-			if (channelSoundContainer->IsBeingPlayed()) {
+			if (channelSoundContainer != nullptr && channelSoundContainer->IsBeingPlayed()) {
 				channelSoundContainer->RemovePlayingChannel(channelIndex);
 			}
 			result = (result == FMOD_OK) ? channel->setUserData(nullptr) : result;
@@ -812,7 +828,8 @@ FMOD_RESULT F_CALLBACK AudioMan::SoundChannelEndedCallback(FMOD_CHANNELCONTROL* 
 			}
 
 			if (result != FMOD_OK) {
-				g_ConsoleMan.PrintString("ERROR: An error occurred when Ending a sound in SoundContainer " + channelSoundContainer->GetPresetName() + ": " + std::string(FMOD_ErrorString(result)));
+				const std::string containerName = channelSoundContainer != nullptr ? channelSoundContainer->GetPresetName() : "<destroyed>";
+				g_ConsoleMan.PrintString("ERROR: An error occurred when Ending a sound in SoundContainer " + containerName + ": " + std::string(FMOD_ErrorString(result)));
 				return result;
 			}
 		} else {
--- a/Source/Managers/AudioMan.cpp
+++ b/Source/Managers/AudioMan.cpp
@@ -603,19 +603,6 @@ bool AudioMan::StopSoundContainerPlayingChannels(SoundContainer* soundContainer,
 	return result == FMOD_OK;
 }
 
-void AudioMan::DisownSoundContainerPlayingChannels(const SoundContainer* soundContainer) {
-	if (!m_AudioEnabled || !soundContainer || !soundContainer->IsBeingPlayed()) {
-		return;
-	}
-	// Leave the channels playing, but null their back-reference so the positional/ended passes don't read the freed container.
-	FMOD::Channel* soundChannel;
-	for (int channelIndex: *soundContainer->GetPlayingChannels()) {
-		if (m_AudioSystem->getChannel(channelIndex, &soundChannel) == FMOD_OK) {
-			soundChannel->setUserData(nullptr);
-		}
-	}
-}
-
 void AudioMan::FadeOutSoundContainerPlayingChannels(SoundContainer* soundContainer, int fadeOutTime) {
 	if (!m_AudioEnabled || !soundContainer || !soundContainer->IsBeingPlayed()) {
 		return;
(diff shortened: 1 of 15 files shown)


external/sources/tracy/public/client/TracyProfiler.cpp
```diff
@@ -1891,7 +1891,7 @@ void Profiler::Worker()
 
                     lastBroadcast = t;
                     const auto ts = std::chrono::duration_cast<std::chrono::seconds>( std::chrono::system_clock::now().time_since_epoch() ).count();
-                    broadcastMsg.activeTime = int32_t( ts - m_epoch );
+                    broadcastMsg.activeTime = ts > m_epoch ? int32_t( ts - m_epoch ) : 0;
```
HeliumAnt: update from upstream instead. Also unnecessary, can't realistically happen until the 2038 bug hits.

meson.build
```diff
@@ -30,10 +30,37 @@ if compiler.get_argument_syntax()== 'gcc' # used for gcc compatible compilers
...
+      '-D_LIBCPP_ENABLE_CXX17_REMOVED_AUTO_PTR',
+      '-D_LIBCPP_ENABLE_CXX17_REMOVED_UNARY_BINARY_FUNCTION',
+      '-D_LIBCPP_ENABLE_CXX17_REMOVED_BINDERS',
+      '-D_LIBCPP_ENABLE_CXX17_REMOVED_RANDOM_SHUFFLE',
+      '-D_LIBCPP_ENABLE_CXX17_REMOVED_UNEXPECTED_FUNCTIONS',
+      '-D_LIBCPP_ENABLE_CXX20_REMOVED_BINDER_TYPEDEFS',
+      '-D_LIBCPP_ENABLE_CXX20_REMOVED_NEGATORS',
+      '-D_LIBCPP_ENABLE_EXPERIMENTAL',
+      '-D_LIBCPP_PSTL_BACKEND_SERIAL',
+      language: 'cpp')
```
HeliumAnt: Remove. Not necessary, already in place in the luabind meson build files. Should also not be global arguments.

Source/Managers/AudioMan.cpp
```diff
@@ -603,6 +603,19 @@ bool AudioMan::StopSoundContainerPlayingChannels(SoundContainer* soundContainer,
...
+		return;
+	}
+	// Leave the channels playing, but null their back-reference so the positional/ended passes don't read the freed container.
+	FMOD::Channel* soundChannel;
+	for (int channelIndex: *soundContainer->GetPlayingChannels()) {
+		if (m_AudioSystem->getChannel(channelIndex, &soundChannel) == FMOD_OK) {
+			soundChannel->setUserData(nullptr);
+		}
+	}
+}
```
HeliumAnt: Hm, this technically works around the crashes ig. Not sure this behaves correctly tho, and will still crash from lua (much rarer, can still happen).

HeliumAnt (changes requested):

Source/Menus/SaveLoadMenuGUI.cpp
```diff
@@ -180,7 +180,13 @@ void SaveLoadMenuGUI::UpdateSaveGamesGUIList() {
 		const auto saveTime = std::chrono::system_clock::to_time_t(saveFsTime);
 #else
 		// TODO - kill this monstrosity when we move to GCC13
+		// macOS libc++ file_clock rep is __int128; duration_cast lands it in system_clock's range first.
+#if defined(__APPLE__) && defined(_LIBCPP_VERSION)
```
HeliumAnt: Applies to libc++ in general, not just Apple

HeliumAnt (changes requested):

Source/Menus/SaveLoadMenuGUI.cpp
HeliumAnt: Does not actually make a difference either, since the time point constructor does this.
Also makes more sense to use a clock_cast here in general, which should also make the code below unnecessary.

Source/Managers/AudioMan.cpp
Madreag: you're right... it'll probably still crash.

The destructor disown covers the C++ side, but `CreateSoundContainer` is `adopt`-bound. When async GC collects one it runs `~SoundContainer` on a threadpool worker while `AudioMan::Update()` reads that channel's `userData` on the main thread. The null checks don't close the window since the container can get freed after the check passes. Also, the looping case looks wrong (disowned `loop -1` channels would keep playing with nothing left to stop them).

I'll remove this from the PR rather than do a partial fix. It's out of scope for the determinism and cross-platform work I should be focusing on.

meson.build
Madreag: Agreed on most of these. cxx17 and cxx20 ones get pulled in through luabind/boost, and nothing outside the lua binding TUs includes boost, so luabind's own args cover them (already in place, like you stated), so I'll remove them all and move the block from `add_global_arguments` to `extra_args` . I agree it shouldn't be global.

`_LIBCPP_ENABLE_EXPERIMENTAL` and `_LIBCPP_PSTL_BACKEND_SERIAL` are needed though. They're not removed-api shims, they're what turns on `<execution>`. libc++ sets `_LIBCPP_HAS_NO_INCOMPLETE_PSTL` and drops the parallel algorithms unless `_LIBCPP_ENABLE_EXPERIMENTAL` is defined, and the engine uses `std::execution::par_unseq` in `PathFinder`, `SLTerrain`, `ActivityMan`, and `SaveLoad`. None of those see luabind's args, so on a libc++ build those TUs won't compile without them. (Upstream CI builds macos through osxcross/GCC, and these don't do anything there, so it's the native clang/libc++ build that needs them.)

Also, did a bit of research on the luabind side:  The `_LIBCPP_ENABLE_CXX17_REMOVED_FEATURES` umbrella it uses got removed in libc++ 19, so on a recent xcode it doesn't do anything anymore.  `_LIBCPP_ENABLE_CXX17_REMOVED_*` macros still work, so if that ever starts breaking the build I could move luabind onto those. That would be another PR though.

Source/Menus/SaveLoadMenuGUI.cpp
Madreag: Agreed, it's all of libc++ (not apple-specific, like you said), so I'll broaden the guard to `#if defined(_LIBCPP_VERSION)`.

The time_point ctor doesn't cover it on libc++ though. file_clock is nanosecond but system_clock is microsecond, so the direct construction won't compile and the duration_cast stays. clock_cast would be cleaner but it doesn't work for file_clock on libc++ yet, so I'll leave the manual path.

external/sources/tracy/public/client/TracyProfiler.cpp
Madreag: I just verified master and v0.13.1 still have the bare `int32_t(ts - m_epoch)` + assert, same as our 0.12.2, so updating wouldn't fix it. It's not really 2038 either. `m_epoch` is `uint64_t`, so a backwards clock step at startup (ntp/vm resync) makes `ts` drop below it and underflow, tripping the assert. Windows shouldn't hit it (NDEBUG compiles it out).

It's a profiler thing so I'll remove it. I could send the guard upstream to tracy separately later.

HeliumAnt: upstream if possible then

meson.build
HeliumAnt: yup, I see. Apple really seems to like weird clang configurations. This literally works just fine on any normal clang build (x86 or arm). Just put it under the clang compiler, since this pollutes the command line quite a bit. And add a comment that it's for apple support, because I will probably forget why it's there otherwise.

Source/Menus/SaveLoadMenuGUI.cpp
HeliumAnt: once again foiled by the strange decisions by apple for the standard library. You are correct, didn't know apples clang distro is so scuffed, doesn't support clock_cast at all. works on all normal llvm clang distributions, just not apple's.

Source/Managers/AudioMan.cpp
HeliumAnt: yeah this is much better addressed in #265

Madreag:
Pushed the changes. The 2 `<execution>` macros sit under a `compiler.get_id() == 'clang'` check now, with a comment on why (meson reports apple clang as plain `clang`, so the gate fires there, checked on the mac). The osxcross gcc CI build stops getting them entirely. Save menu guard is `#if defined(_LIBCPP_VERSION)`, and Tracy and the AudioMan/SoundContainer changes are out. I'll send the tracy guard upstream separately.

One correction to what I said though... dropping the root macros broke luabind on mac, xcode is already on libc++ 20 where the `_LIBCPP_ENABLE_CXX17_REMOVED_FEATURES` umbrella does nothing, so its TUs stopped compiling on `std::auto_ptr`. The swap to the individual macros couldn't wait for another PR, it's in here as its own commit.

Also removed two stale LuaJIT changelog bullets that snuck in from a rebase (already merged in #280). Rebuilt and booted on Windows and the arm64 mac (apple clang 17 / libc++ 20).

HeliumAnt (approved):
Looks good now. I'll also make sure to update the CI to also build macOS ARM and apple clang asap.

Causeless (approved):

Causeless:
Nice one!

PR #284, opened 2026-09-13
fix: make spectator camera dependency self-contained

mythzilla-coder:
## Summary
- package the camera event helper and pure test required by SpectatorArena
- add a dated acceptance report recording HOLD_FOR_VISUAL_ACCEPTANCE
- keep production AI_V2_MODE=OFF and preserve existing evidence artifacts

## Verification
- python -m unittest discover -s tests -p '*.py' -v (12 passed)
- Fengari camera/controller/telemetry tests pass
- Debug Release x64 MSBuild exits 0 (existing warnings remain)
- git diff --check passes

## Review status
Camera behavior is intentionally not accepted yet: native rendered-frame review could not be completed because the native computer-use surface exposed no game window. See `docs/SPECTATOR_CAMERA_ACCEPTANCE_2026-09-13.md`.


Causeless:
Sorry, I'll need some sort of explanation of what this is and what it adds. Is this is a spectator camera gamemode, I believe that there are already mods that achieve this with much less code bloat.

mythzilla-coder:
Thanks for taking a look. Your concern about the size of the change is fair.

What we're working on is broader than a simple spectator-camera mod.
The goal is an autonomous AI-vs-AI spectator mode that can run
continuous matches by itself, handle scoring/resetting, select and
attribute important combat events, move the camera toward those
events, and eventually serve as a stable spectator/livestream mode.

A lot of the code currently visible in this PR is diagnostic and
experimental infrastructure that accumulated while we were
investigating engine behavior — for example actor lifecycle/death
attribution, camera timing, telemetry, and runtime verification. That
makes the PR much larger than the actual production feature should be.

In hindsight, we opened the PR too early and then used the branch as
an active R&D workspace. That's not a good fit for upstream review,
and I don't want to burden the project with a large experimental diff.

I'm going to close this PR and continue the work privately. If we
later identify a small, self-contained improvement that genuinely
belongs upstream, I'll come back with a much cleaner and more focused
contribution.

Thanks for the feedback — it helped clarify that distinction.

